PR #47, opened 2014-12-12
Allow multiple config files

fxposter:
"-config" option now accepts wildcards (ie: it's processed via
go's `filepath.Glob`) and multiple files with ",":

    docker-gen -config /example/a,/example/b,/usr/configs/*

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,8 +79,9 @@ Usage: docker-gen [-config file] [-watch=false] [-notify="restart xyz"] [-notify
 *Options:*
 
 ```
-  -config="": Use the specified config file instead of command-line options.  Multiple templates can be defined and
-              they will be executed in the order that they appear in the config file.
+  -config="": Use the specified config file[s], separated by comas, instead of command-line options. All file patterns
+              are processed with [filepath.Glob](http://golang.org/pkg/path/filepath/#Glob). One file can contain
+              multiple templates and they will be executed in the order that they appear in the config file.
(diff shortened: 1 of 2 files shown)


md5:
I'd prefer to see `docker-gen` accept multiple `-config` settings each with their own config file. If you really want to be able to use globbing, it could make sense for the shell to do the globbing and have `-config` take multiple arguments (greedy up until the next `-param`).

fxposter:
Actually, globbing is the first step in the direction I'd like to have - so that multiple application provide templates and configs to single docker-gen instance and so that docker-gen automatically catch file changes and filelist changes and regenerate templates based on new configs/

md5:
I'm just not a fan of utilities doing their own glob processing of command-line arguments if it can be avoided.

As for `docker-gen` regenerating templates based on new configs, I don't think you could do that with globs anyways since `docker-gen` doesn't watch config files for changes. If a directory-based approach were taken, `docker-gen` could potentially use `inotify` to watch for config file changes (at least on Linux), but there no way to register an inotify watch for a glob that I'm aware of.

fxposter:
You could just poll on selected interval for anything ;)

md5:
I just opened an alternate PR at #84 that supports passing the `-config` flag multiple times to specify multiple config files. I didn't add the globbing because of the complications I mentioned before, but I think the globbing is orthogonal to supporting multiple config files.

jwilder:
Closing this in favor of #84.

Issue #26, opened 2014-09-22
Add a way to filter out containers in templates


vlcinsky:
Just following the last sentence from README (TODO).

Current set of functions allowed in templates is very limited and does not allow much filtering.

Even subset from following functions would help:

- equality of strings
- test, if a string is present in an array of string (for this we need also a function to define in template the array of string)
- wildcard matching
- startsWith

dataferret:
I'd be interested in something like where too. Although intersect would be nice too for those moments where we have a comma list of filter tags and a comma list of env tags.  Any work so far on bringing some of the Hugo stuff over?

jwilder:
Closing this w/ since #54 has landed.  Equality/prefix/suffix and a few where functions should provide better ability to filter containers in templates.  We can open new issues/PR for other template functions that would be useful.

PR #54, opened 2015-01-28
Where primitives

dataferret:
Related to Issue #26. I added intersect (slice vs slice compare), and a few where functions that utilize it to match.


md5:
:+1:

md5:
@dataferret just saw your comment on #26. I just pushed [an old commit](https://github.com/md5/docker-gen/commit/f6144b8545657f0eb6f5ed90607c658f1bf87daa) I had locally for using `Where` from `hugolib`, but I think just implementing it directly like you did is better.

jwilder:
This is nice.  What do you think about some slightly different names?  

* `whereSomeMatch` -> `whereAny`
* `whereRequires` -> `whereAll`

dataferret:
Those names sound good.  You want to do it your end or should I resubmit the PR?

jwilder:
Merged w/ bcddaab30.  Thanks!

PR #55, opened 2015-01-28
Add Travis CI configuration

md5:
This PR adds configuration for [Travis CI](https://travis-ci.org).

The tests on [my Travis setup](https://travis-ci.org/md5/docker-gen/builds/48577781) are currently failing, but that's because `getEndpoint` is calling `exists` on `/var/run/docker.sock` and Docker isn't running in the Travis environment. I have a similar issue with that same test failing locally because I'm running boot2docker and my `DOCKER_HOST` is set to point to my b2d VM.

Travis would still need to be turned on for `jwilder/docker-gen` for this to work.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,9 @@
+language: go
+go:
+- 1.3
+- 1.4
+- tip
+install:
+- make get-deps
+script:
+- make all test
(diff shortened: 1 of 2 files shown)


dataferret:
I think you might be able to move the (3?) getEndpoint related tests into their own file with `// +build !travis` as a guard?

md5:
I'd actually prefer getEndpoint to not call exists. It doesn't do any
similar checks on a tcp:// DOCKER_HOST.

jwilder:
Cool.  I've turned it on for `jwilder/docker-gen`.  We can probably move the exist check out of the getEndpoint.  It might not even be necessary since trying to open it will still fail and should already be handled.

md5:
Yeah, I think if it isn't there, then a connection to a missing Unix socket will fail at the same time as a bad TCP connection. That seems appropriate to me.

md5:
I also just realized that I might have added the build status badge to `README.md`. The markdown looks like this

```Markdown
[![Build Status](https://travis-ci.org/jwilder/docker-gen.svg?branch=master)](https://travis-ci.org/jwilder/docker-gen)
```

jwilder:
Added

PR #56, opened 2015-01-28
Remove exists() from getEndpoint()

jwilder:
Should fix tests running under travis build and boot2docker where
unix:///var/run/docker.sock does not exist.

exists was not necessary since connecting to a non-existent unix
socket would fail anyways and is already handled.


md5:
:metal:

PR #60, opened 2015-02-23
Added whereExist and whereNotExist

ejhayes:
This includes template helpers `whereExist` and `whereNotExist`.  I found myself looking for containers that did not have an environment variable and wanted an easy way to query containers (existing `where` function didn't seem to work with `""` or `nil`).


thaJeztah:
I think there's an error in the Dockerfile, because it now refers to *your* repository, not the "main" repository.

(Perhaps the Dockerfile changes should not be part of the pull-request)

ejhayes:
Yikes, let me remove that

ejhayes:
Alright removed - thanks for pointing that out

thaJeztah:
Hehe, no sweat, easy mistake to make, I just spotted it :)

jwilder:
Thanks!

ejhayes:
Thanks @jwilder - want to add this to `0.3.7` or `0.3.8`?

md5:
@ejhayes I just thought I'd point out that Git tags should generally not be changed after they're created, so the thing to do would be to cut a new version.

ejhayes:
@md5 totally makes sense. This should be in `0.3.8` in then :)

md5:
:metal:

jwilder:
Yeah... will go in 0.3.8.

PR #61, opened 2015-02-26
Run go fmt

md5:
I'm working on some other changes, but I figured I'd send through this commit.

@jwilder would you be in favor of changing the Travis build to fail if `go fmt` hasn't been run?


jwilder:
@md5 Yeah. Sounds good.

md5:
cf. #62

PR #63, opened 2015-02-26
Separate newTemplate from generateFile

md5:
I originally wrote this code as the start of an effort to allow sub-template includes, but I'm pushing it as a separate PR since it's useful for testing (see included test update for `TestKeys`).


gingerlime:
@md5 just curious if anything happened with the idea of sub-template includes?

PR #65, opened 2015-02-28
Fix code for splitting KEY=VALUE env vars

md5:
Make `docker-gen` work with env values with `=` in them as well as entries of the form `KEY` (no `=`)

As reported in jwilder/nginx-proxy#110 and zettio/weave#422

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,7 +25,7 @@ dist: dist-clean
 
 
 release: dist
-	glock sync github.com/jwilder/docker-gen
+	glock sync -n < GLOCKFILE
 	tar -cvzf docker-gen-linux-amd64-$(TAG).tar.gz -C dist/linux/amd64 docker-gen
 	tar -cvzf docker-gen-linux-i386-$(TAG).tar.gz -C dist/linux/i386 docker-gen
 	tar -cvzf docker-gen-linux-armel-$(TAG).tar.gz -C dist/linux/armel docker-gen
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 .SILENT :
 .PHONY : docker-gen clean fmt
 
-TAG:=`git describe --abbrev=0 --tags`
+TAG:=`git describe --tags`
 LDFLAGS:=-X main.buildVersion $(TAG)
 
 all: docker-gen
--- a/docker-gen.go
+++ b/docker-gen.go
@@ -100,8 +100,11 @@ type Context []*RuntimeContainer
 func (c *Context) Env() map[string]string {
 
 	env := make(map[string]string)
-	for _, i := range os.Environ() {
-		parts := strings.Split(i, "=")
+	for _, entry := range os.Environ() {
+		parts := strings.SplitN(entry, "=", 2)
+		if len(parts) != 2 {
+			parts = append(parts, "")
+		}
 		env[parts[0]] = parts[1]
 	}
 	return env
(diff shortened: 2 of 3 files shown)


setiseta:
thx for the fast try to fix this.
i've tested this with lokal build and passed the new docker-gen into jwilder/nginx-proxy with -v
but it seems not to fix this problem.
i'm not familiar with go, so i don't know how to debug or get more information.

i just got this:
root@2e0f48c2694e:/app# forego start -r
forego     | starting nginx.1 on port 5000
forego     | starting dockergen.1 on port 5100
dockergen.1 | panic: runtime error: index out of range
dockergen.1 |
forego     | starting dockergen.1 on port 5200
forego     | sending SIGTERM to dockergen.1
forego     | sending SIGTERM to nginx.1

and if i run the docker-gen like in the Procfile i get this:
root@2e0f48c2694e:/app# docker-gen -watch -only-exposed -notify "nginx -s reload" /app/nginx.tmpl /etc/nginx/conf.d/default.conf
panic: runtime error: index out of range

goroutine 16 [running]:
runtime.panic(0x796a80, 0x9ec79c)
        /usr/lib/golang/src/pkg/runtime/panic.c:279 +0xf5
main.getContainers(0xc20804c200, 0x0, 0x0, 0x0, 0x0, 0x0)
        /pool/build/docker-gen/docker_client.go:152 +0xcb3
main.generateFromContainers(0xc20804c200)
        /pool/build/docker-gen/docker-gen.go:157 +0x35
main.main()
        /pool/build/docker-gen/docker-gen.go:395 +0x40c

goroutine 19 [finalizer wait]:
runtime.park(0x41e620, 0x9f0808, 0x9eea49)
        /usr/lib/golang/src/pkg/runtime/proc.c:1369 +0x89
runtime.parkunlock(0x9f0808, 0x9eea49)
        /usr/lib/golang/src/pkg/runtime/proc.c:1385 +0x3b
runfinq()
        /usr/lib/golang/src/pkg/runtime/mgc0.c:2644 +0xcf
runtime.goexit()
        /usr/lib/golang/src/pkg/runtime/proc.c:1445

if you can give me some advises i can give you more information.

md5:
@jwilder LMK what you think about the `Makefile` commits. I can remove them from this branch before merging if you thing those changes need to be discussed separately or removed entirely.

md5:
After sleeping on this, I had a few thoughts.

I think that the `strings.SplitN` changes should go in no matter what. However, it looks like `os.Environ()` may guarantee an "=", so the `append` may not be necessary there.

In the case of the `KEY` environment variables, this is a case where Docker's env-passthrough functionality has been used with `-e KEY`, but `$KEY` is not defined in the invoking environment. If you run `env` inside such a container, `KEY` is not included in the output:

```
$ env KEY2= docker run --rm -e KEY1 -e KEY2 -e KEY3= debian:jessie env
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
HOSTNAME=57347b4b5aea
KEY2=
KEY3=
HOME=/root
```

Because of this, it may be best to omit such variables in `getContainers` instead of including them with a value of `""`.

jwilder:
I amended this by extracting that splitting code into a utils func and added some tests.  

Thanks @md5

md5:
Cool. I still think it might be best to leave out the `KEY` entries with no `=`, but it's a minor issue.

md5:
Looking forward to an update to `nginx-proxy` that includes these changes since that means we'll get 'parseJson` too :)

Issue #50, opened 2014-12-21
unable to parse template: template: docker-gen:2208: unexpected unrecognized character in action: U+0000 in command


udono:
Hi,
I am not sure if this issue is related to docker-gen.
When I try to run your examples with http://fig.sh I get the following error message:

    dockergen_1 | 2014/12/21 14:14:37 unable to parse template: template: docker-gen:2208: 
    unexpected unrecognized character in action: U+0000 in command

My fig.yml looks like this:

        nginx:
          image: nginx
          volumes:
            - /tmp/nginx:/etc/nginx/conf.d
          ports:
            - "80:80"

        dockergen:
          image: jwilder/docker-gen:0.3.6
          command: /usr/local/bin/docker-gen -notify-sighup nginx -watch --only-published /etc/docker-gen/templates/nginx.tmpl /etc/nginx/conf.d/default.conf
          volumes:
            - /var/run/docker.sock:/tmp/docker.sock
            - /tmp/templates:/etc/docker-gen/templates
          volumes_from:
            - nginx

Ideas?

andreaskoch:
Hi,
I encountered exactly the same error but I found the solution.

The reason by docker-gen does not work is because the [entrypoint](https://github.com/jwilder/docker-gen/blob/master/Dockerfile#L10) of the `jwilder/docker-gen` image is already set to `docker-gen`, so you basically pass `/usr/local/bin/docker-gen` as the first argument to it and thats why it isn't working.

If you modify your Fig file like this it will work:

```yaml
nginx:
  image: nginx
  volumes:
    - /tmp/nginx:/etc/nginx/conf.d
  ports:
    - "80:80"

dockergen:
  image: jwilder/docker-gen:0.3.6
  command: -notify-sighup nginx -watch --only-published /etc/docker-gen/templates/nginx.tmpl /etc/nginx/conf.d/default.conf
  volumes:
    - /var/run/docker.sock:/tmp/docker.sock
    - /tmp/templates:/etc/docker-gen/templates
  volumes_from:
    - nginx
```

@jwilder: Thanks for this great tool.

Greetings
Andy

jwilder:
Thanks @andreaskoch!

swapnil7434:
Hi,

I am getting following error.

[nginx-gen-1]2017-07-31T06:36:52.861965607Z 2017/07/31 06:36:52 Unable to parse template: open /etc/docker-gen/templates/nginx.tmpl: no such file or directory

[...]

PR #66, opened 2015-03-01
Optimize Docker image

md5:
I ended up looking at the Dockerfile for docker-gen after seeing this comment: https://github.com/jwilder/docker-gen/issues/50#issuecomment-70714517

I made the following improvements to help speed up and optimize the Docker image build:

* Remove and unpack TAR file in the same layer as the download
* Clean up more leftovers from installing curl
* Add .dockerignore to minimize build context to necessary files

I actually think that the `ENTRYPOINT` should be changed to a `CMD` too, but that would be a breaking change for some users and I didn't include it.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,4 @@
+.git
+docker-gen
+dist
+*.gz
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,9 +2,20 @@ FROM debian:wheezy
 MAINTAINER Jason Wilder <jwilder@litl.com>
 
 ENV VERSION 0.3.7
+ENV DOWNLOAD_URL https://github.com/jwilder/docker-gen/releases/download/$VERSION/docker-gen-linux-amd64-$VERSION.tar.gz
 ENV DOCKER_HOST unix:///tmp/docker.sock
 
-RUN apt-get update && apt-get install -y curl && curl -o docker-gen-linux-amd64-$VERSION.tar.gz -L https://github.com/jwilder/docker-gen/releases/download/$VERSION/docker-gen-linux-amd64-$VERSION.tar.gz && apt-get remove -y curl && apt-get -y clean
-RUN tar -C /usr/local/bin -xvzf docker-gen-linux-amd64-$VERSION.tar.gz && rm docker-gen-linux-amd64-$VERSION.tar.gz
+RUN deps=' \
+		curl ca-certificates \
+	'; \
+	set -x; \
+	apt-get update \
+	&& apt-get install -y --no-install-recommends $deps \
+	&& curl -o docker-gen.tar.gz -L $DOWNLOAD_URL \
+	&& tar -C /usr/local/bin -xvzf docker-gen.tar.gz \
+	&& rm docker-gen.tar.gz \
+	&& apt-get purge -y --auto-remove -o APT::AutoRemove::RecommendsImportant=false -o APT::AutoRemove::SuggestsImportant=false $deps \
+	&& apt-get clean -y \
+	&& rm -rf /var/lib/apt/lists/*
 
 ENTRYPOINT ["/usr/local/bin/docker-gen"]


thaJeztah:
> I actually think that the ENTRYPOINT should be changed to a CMD too, but that would be a breaking change for some users and I didn't include it.

The advantage of having it as "entrypoint", is that the whole image can be used as it it was the 'docker-gen', so `docker run --rm jwilder/docker-gen --version`. It could even be set as an `alias`

Two thoughts to get the image-size down even further (and/or simplify the Dockerfile);

1. Would it be possible to put raw binaries on the releases page (no .tar.gz)? That would allow skipping the whole curl/untar steps, and simply do:

    ADD https://github.com/jwilder/docker-gen/releases/download/$VERSION/docker-gen-linux-amd64-$VERSION.tar.gz /usr/local/bin/docker-gen

2. Is it possible to build docker-gen statically? This would make it possible to have an image containing *just* the binary. For reference, the official Docker Swarm image (https://github.com/docker/swarm-library-image/blob/d9999a8a797d8e9744f1bfc0e7211d016478b2fc/Dockerfile) is 7mb total.

md5:
I just consider this the low-hanging fruit based on the current Dockerfile. It would be interesting to play around with getting a really small image based on `scratch` like Docker Swarm does, it's not something I've been inclined to do myself.

md5:
@thaJeztah do you know how they're building the `swarm` binary? I don't see a Makefile in https://github.com/docker/swarm

thaJeztah:
@md5 I'm :+1: on the changes in this PR, but it got me thinking if more awesomeness was possible

> do you know how they're building the swarm binary? I don't see a Makefile in

Hm, I think I saw it pass on IRC when it was first added to the official library. I can try and find the logs (in #docker-library), or you can ask "aluzzardi" (left out the `@` to not ping him unnecessarily)

thaJeztah:
@md5 oh, think I found it; https://github.com/docker-library/official-images/pull/440#issuecomment-73075439

md5:
@thaJeztah I remember following that discussion, but as you say it doesn't say how they're actually building `swarm`.

After a bit of Googling, I found a few blog posts about how to create a statically compiled Go binary and created this: https://github.com/md5/docker-gen/tree/static-build

This was the result:

```
$ docker images docker-gen
REPOSITORY          TAG                 IMAGE ID            CREATED             VIRTUAL SIZE
docker-gen          latest              26cd1e51a52f        3 minutes ago       6.746 MB
docker-gen          build               7aefbbca8031        3 minutes ago       557.8 MB
```

I haven't had a chance to play around with it yet and probably won't get to today.

thaJeztah:
:100: Oh, that's looking awesome! I'll see if I can find some time to test it as well (but it's quite hectic at work atm, so no promises)

Are there known use-cases of people building *on top of* the docker-gen image? If that's the case, I can imagine offering both variations (debian + docker-gen and only "bare" docker-gen). If not, making this the default does look good.

md5:
I don't know of anyone building on top of the `docker-gen` image, but who knows if that's in the wild. I definitely think the way to go is with multiple tags, though getting an automated build for this setup is going to be tricky (someone or something would still have to build and push a `docker-gen` binary to Github to trigger the automated build, possibly from a branch or something).

I did get a minute to test this and it appears to work fine (see the tip of the `static-build` branch since I added another commit):

```
$ docker run --rm -v /var/run/docker.sock:/tmp/docker.sock -v $(pwd)/templates:/templates:ro -w /templates -v /output docker-gen nginx.tmpl /output/nginx.conf
2015/03/01 20:57:19 Generated '/output/nginx.conf' from 4 containers
```

I checked the output and it looks as it should.

md5:
I just did a quick search for people making images `FROM jwilder/docker-gen`: https://github.com/search?l=dockerfile&q=from+docker-gen&ref=searchresults&type=Code&utf8=%E2%9C%93

There's at least one result from the first page: https://github.com/nuclyus/docker-juhasz

md5:
I took a closer look at how Docker Swarm is building their binary and it looks like they actually do it in the `Dockerfile` of the `docker/swarm` repo itself. The normal build there is based on `golang:1.3` and builds a static `swarm` binary that it uses as the `ENTRYPOINT` as the image. There is then an `update.sh` script over in `docker/swarm-library-image` that is responsible for getting the binary out of the first image with `docker cp` as well as for extracting a CA root bundle from `debian:latest`.

I've made a couple changes to the [`static-build`](https://github.com/jwilder/docker-gen/compare/master...md5:static-build) branch to merge `Dockerfile` and `Dockerfile.build` and created an orphan branch at [`static-build-image`](https://github.com/md5/docker-gen/tree/static-build-image) as a POC of the separated image build.

md5:
Apparently the `apt-get clean -y` isn't necessary due to some [voodoo magic](https://github.com/docker/docker/blob/1aa67071cb3e6bc8a08f662d913cdb1b1eda1645/contrib/mkimage/debootstrap#L79-80) in Docker.

thaJeztah:
^^ just read the discussion there, talk about bad UX 😞

md5:
Ping @jwilder

PR #71, opened 2015-03-10
Include information about the swarm node a container is on.

JustAdam:


md5:
@JustAdam I think you might need to update docker/docker to the same ref that fsouza/go-dockerclient has in order to fix that Travis failure.

JustAdam:
Thanks @md5 got that fixed with the latest commit from Docker.

md5:
:+1:

jwilder:
Thanks!

PR #76, opened 2015-03-17
Set host IP on container addresses

csabapalfi:
This is especially useful when using docker-gen with swarm.

Could also use the new Node object but this felt simpler.


csabapalfi:
Aah, just saw what you did in 0.3.9. My PR is no longer relevant I guess.

Issue #81, opened 2015-04-20
Use labels instead of environment variables 


mariusGundersen:
The latest version of docker let's you label a container with metadata, associating a string value to a string key. This seems like a better fit for VIRTUAL_HOST than using environment variables

thaJeztah:
I think both can make sense. One advantage of env is that they are accessible inside the container, which can be handy.

Nonetheless, adding support for labels could be nice :)

md5:
I believe this can be closed now.

jwilder:
Fixed by #82

PR #82, opened 2015-05-02
Expose RuntimeContainer.Labels

md5:
Related: jwilder/docker-gen#81

### DIFF
--- a/GLOCKFILE
+++ b/GLOCKFILE
@@ -1,4 +1,4 @@
 github.com/BurntSushi/toml f87ce853111478914f0bcffa34d43a93643e6eda
 github.com/Sirupsen/logrus 6ebb4e7b3c24b9fef150d7693e728cb1ebadf1f5
 github.com/docker/docker 95dcb2748442fc1f59e8c2e010679e0aea3cef1c
-github.com/fsouza/go-dockerclient de7e7a55fa1ff99c88420e104e4966f9ad36c72d
+github.com/fsouza/go-dockerclient 1e60afa8d58b91a0eca0a732927c8ecba696493f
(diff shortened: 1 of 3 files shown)


jwilder:
@md5 nice! Been meaning to add this but haven't had time.  Need to test it out but looks pretty straightforward.

md5:
:+1:

thaJeztah:
Great!


*/me should suggest to add @md5 as maintainer to this and the nginx-proxy repos*

tompson:
Is there a sample somewhere on how to use labels in a template? I somehow can not get it to work

tompson:
The problem is that the labels generated by compose are looking like 

```com.docker.compose.service```

How do you access them in the template?

**Comment by md5** [comment]
I believe `index .Env.Labels "com.docker.compose.service"` should work.

**Comment by trnubo** [comment]
Sorry if this is the wrong place to ask. How would you use a label such as "com.docker.compose.service" in as part of a $fieldPath with a groupBy or where?

`where . "Labels.com.docker.compose.service" "web"`

**Comment by md5** [comment]
@trnubo I don't think it's going to be possible to do that right now. I also think it would be tricky to get right.

Adding `groupByLabel` and `whereLabel` functions might not be the worst idea, although it would balloon pretty quickly if you add `*Label` variants of all the existing `where*` and `groupBy*` functions.

**Comment by md5** [comment]
@trnubo Take a look at https://github.com/jwilder/docker-gen/pull/143

**Comment by md5** [comment]
For the case you mentioned, you could do this:

```
whereLabelValueMatches . "com.docker.compose.service" "^web$"
```

**Comment by helderco** [comment]
What's wrong here?

```
{{ range $host, $containers := groupByMulti $ "Env.VIRTUAL_HOST" "," }}
{{ $project := index $.Labels "com.docker.compose.project" }}
...
```

I get this error: `template error: template: nginx.tmpl:61:22: executing "nginx.tmpl" at <$.Labels>: can't evaluate field Labels in type *dockergen.Context`

PR #83, opened 2015-05-03
Generalize where* functions

md5:
This PR generalizes the `where*` functions to allow an array or slice of any map or struct type.

This should make it possible to fix https://github.com/jwilder/nginx-proxy/issues/115 since the updated `where` function can be used like this: `{{where (where $container.Addresses "Proto" "tcp") "Port" "80" | len}}` to detect the case where there are multiple ports exposed but not port 80.

I also introduced a `templateTestList` type with a `run` function to simplify writing tests that actually execute a template and check its output. Let me know if you'd like that split into a separate PR.


md5:
Looks like `os.Unsetenv` is only in Go 1.4+, causing the Travis failure. I've changed it to `os.Setenv("DOCKER_HOST","")`

thaJeztah:
Are there blockers for Go 1.4 in docker-gen, or are people importing it in other projects? Perhaps 1.4 isn't a bad thing to use (Docker, for example, very recently dropped 1.3 support)

md5:
Good point @thaJeztah.

jwilder:
I don't see any reason to keep 1.3 around.  `docker-gen` isn't really intended to be used as a library right now so it should not matter if the code is dependent on 1.4 features.

md5:
@jwilder I've dropped commit 844fa876c2f65a79ed5e05dfe06f2c256e345b07, so the Travis build now fails again on 1.3.

jwilder:
@md5 I removed 1.3 from travis.  If you rebase, it should pass now.

md5:
Rebased. Travis is green :+1:

jwilder:
Thanks @md5!

md5:
You're welcome! 👍

PR #88, opened 2015-05-19
broke out the filters into functions

jasonwbarnett:
I did this so it would be easier to add additional filters in the future (which I plan to do).

### DIFF
--- a/template.go
+++ b/template.go
@@ -336,22 +336,10 @@ func generateFile(config Config, containers Context) bool {
 		log.Fatalf("unable to parse template: %s", err)
 	}
 
-	filteredContainers := Context{}
-	if config.OnlyPublished {
-		for _, container := range containers {
-			if len(container.PublishedAddresses()) > 0 {
-				filteredContainers = append(filteredContainers, container)
-			}
-		}
-	} else if config.OnlyExposed {
-		for _, container := range containers {
-			if len(container.Addresses) > 0 {
-				filteredContainers = append(filteredContainers, container)
-			}
-		}
-	} else {
-		filteredContainers = containers
-	}
+	// Pass containers through our filters
+	filteredContainers := containers
+	filterOnlyPublished(&config, &filteredContainers)
+	filterOnlyExposed(&config, &filteredContainers)
 
 	dest := os.Stdout
 	if config.Dest != "" {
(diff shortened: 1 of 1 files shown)


md5:
Methinks this diff would be much easier to read if you put the helper functions below `generateFile` :+1:

jasonwbarnett:
updated

md5:
Thanks for updating the diff!

I hate to say it, but the `removeFilteredContainer` stuff seems too clever to me. You also changed the logic around how `OnlyPublished` and `OnlyExposed` were mutually exclusive, though this shouldn't be a problem in practice since "published" is always a subset of "exposed".

I think in terms of extensibility, I'd prefer to see something like a `Filters` field on `Config` that contains a list of `type Filter func(*RunningContainer) bool` functions. I'd use a single pass over the container list and iterate through `Filters` to make sure that every filter passed for the container.

I suppose this assumes that all the filters anyone would be interested in would be those that look at a single container. I could imagine someone wanting a filter that has all the other containers for context, but I can't think of an example. My idea also would mean that the config file unmarshaling would become more complicated. (I think I just talked myself out of it...)

Let's see what @jwilder has to say, since my opinion here is not official :+1:

jasonwbarnett:
1. I appreciate __all__ of the feedback.
2. I must admit, `removeFilteredContainer` is pretty clever :wink:. I can add some additional comments if you think that it will be easier to read/understand for fresh eyes. Let me know.
3. I thought about the passing all filters, but felt like it was way too much initially.
4. I would view this PR as progress towards "cleaner" code. smaller functions == cleaner code, in my opinion. So don't view these changes as permanent changes in any particular directions :smile:

template.go
```diff
@@ -393,10 +381,52 @@ func generateFile(config Config, containers Context) bool {
...
+
+/*
+The idea behind this function is to easily remove a RuntimeContainer pointer
+from the Context slice.
+
+This is intended to be used with a for loop and to pass the "index" as a pointer
+so that after we remove the particular RuntimeContainer pointer, we can bump
+back the index so we don't skip over any of the RuntimeContainer pointers.
+*/
+func removeFilteredContainer(containers *Context, index *int) {
```
jwilder: This is seems more complicated than necessary.  I'd rather the filter functions just return a new filtered slice as opposed to modifying the parameter pointers.

jwilder:
I think pulling these out into separate functions is good but I agree w/ @md5 that the implementation is bit too clever.

jasonwbarnett:
@jwilder do you have any suggestions? I did it that way because in my opinion it was the simplest (I know, crazy, right?). My background is Ruby and I just learned Go 2 days ago, so I have no idea of what a simpler, but not way more code alternative would be. :smile:

jasonwbarnett:
I just saw your comment... I will update the PR.

template.go
```diff
@@ -393,10 +381,52 @@ func generateFile(config Config, containers Context) bool {
...
+	   *containers = c
+
+	   *index--
+	*/
+
+	*containers = append((*containers)[:*index], (*containers)[*index+1:]...)
+	*index--
+}
+
+func filterOnlyPublished(config *Config, containers *Context) {
```
jwilder: I'd suggest something like:

```
func filterOnlyPublished(containers Context) Context {
	filtered := Context{}
	for _, container := range containers {
		if len(container.PublishedAddresses()) > 0 {
			filtered = append(filtered, container)
		}
	}
	return filtered
}
```

And call it like:
```
if config.OnlyPublished {
	filteredContainers = filterOnlyPublished(containers)
} 
```

jasonwbarnett:
This entire thing started out of my personal requirement to add many more filters. My thinking was to put everything required for a filter into it's function. Can we compromise and do as you've suggested, but move the if statement into the `filterOnlyPublished ` function?

jasonwbarnett:
I've decided to abandon this PR. I can't get on board with what you're asking because I believe it is hard to follow and I honestly can't even get my brain to think in this direction. My bad, not yours.

I do understand where you are coming from in regards to the `removeFilteredContainer ` being "too clever", but I do believe it makes more sense overall, at least in the short term until a new requirement comes along.

Unfortunately, we must agree to disagree.

P.S. Thank you for writing docker-gen, it is super awesome and I will definitely use it.

jasonwbarnett:
This is what I was trying to get to https://github.com/jwilder/docker-gen/compare/master...jasonwbarnett:master

PR #90, opened 2015-06-10
Add VIRTUAL_WEIGHT for load balancing

kelunik:
Add an easy way to load balance requests with the nginx proxy template.


uptownhr:
This would be an awesome addition

jwilder:
This template is just an example.  Was this intended for `nginx-proxy`?

kelunik:
Yes, should I open a PR there or do you want to do that?

Issue #73, opened 2015-03-11
Notify HUP not working from config file


jwilder:
There appears to be no way to configure a `-notify-sighup` option via a config file.  This should be possible.

emsi:
+1 on this :)

alefi87:
+1

emsi:
Actually it even works without the patch. :)
All you need to do is use proper syntax in template file as shown in the 6f32cc3f4c6c398028a6214abf4baf26a8a2aa52

Or check:
https://raw.githubusercontent.com/emsi/Docker/master/docker-gen/docker-gen.cfg
and 
https://github.com/emsi/Docker/tree/master/docker-gen

jwilder:
Thanks @emsi!

PR #92, opened 2015-06-16
A fix for bug #73 

emsi:
I made a fix for bug #73 with usage example.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@ MAINTAINER Jason Wilder <jason@influxdb.com>
 
 ENV VERSION 0.4.0
 ENV DOWNLOAD_URL https://github.com/jwilder/docker-gen/releases/download/$VERSION/docker-gen-linux-amd64-$VERSION.tar.gz
-ENV DOCKER_HOST unix:///tmp/docker.sock
+ENV DOCKER_HOST unix:///var/run/docker.sock
 
 RUN deps=' \
 		curl ca-certificates \
--- a/examples/docker-gen.cfg
+++ b/examples/docker-gen.cfg
@@ -8,3 +8,11 @@ notifycmd = "/etc/init.d/nginx reload"
 template = "/etc/logrotate.conf.tmpl"
 dest = "/etc/logrotate.d/docker"
 watch = true
+
+[[config]]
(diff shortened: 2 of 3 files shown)


emsi:
@jwilder: Could you review it please?

examples/docker-gen.cfg
```diff
@@ -8,3 +8,11 @@ notifycmd = "/etc/init.d/nginx reload"
 template = "/etc/logrotate.conf.tmpl"
 dest = "/etc/logrotate.d/docker"
 watch = true
+
+[[config]]
+Template = "/etc/docker-gen/templates/nginx.tmpl"
+Dest = "/etc/nginx/conf.d/default.conf"
+watch = true
+[config.NotifyContainers]
+nginx = 1  # 1 is a signal number to be sent; here SIGINT
```
jwilder: I think this would be more self-documenting and easier to configure if the signal was a string.  e.g.

```
nginx = "SIGINT"
```

`docker-gen` can map the string to the signal number.  I would also make the string comparison case-insensitive too.

emsi: Honestly I'm used to kill -1 rather than kill -SIGINT so 1 seemed more logical to me :)

Dockerfile
```diff
@@ -3,7 +3,7 @@ MAINTAINER Jason Wilder <jason@influxdb.com>
 
 ENV VERSION 0.4.0
 ENV DOWNLOAD_URL https://github.com/jwilder/docker-gen/releases/download/$VERSION/docker-gen-linux-amd64-$VERSION.tar.gz
-ENV DOCKER_HOST unix:///tmp/docker.sock
+ENV DOCKER_HOST unix:///var/run/docker.sock
```
md5: Isn't this going to break everyone who currently invokes their `docker-gen` container with `-v /var/run/docker.sock:/tmp/docker.sock`?

emsi: Yes it is.
The fix was at 6f32cc3f4c6c398028a6214abf4baf26a8a2aa52 which does not include that change.

jwilder:
Seems this was just a documentation issue and doesn't require a code fix.  Fixed via 39f07.

PR #97, opened 2015-07-10
generalizedGroupBy with reflection

webner:
the release 0.4.0 breaks following template because "where" returns now a generic interface{} and "groupBy" expects []*RuntimeContainer

```
{{ $containers := where $ "Env.Filter" "true" }}
{{ range $host, $containers := groupByMulti $containers "Env.VIRTUAL_HOST" "," }}
```

This PR changes the groupBy arguments also to interface{}

### DIFF
--- a/template.go
+++ b/template.go
@@ -29,52 +29,84 @@ func exists(path string) (bool, error) {
 	return false, err
 }
 
-func groupByMulti(entries []*RuntimeContainer, key, sep string) map[string][]*RuntimeContainer {
-	groups := make(map[string][]*RuntimeContainer)
-	for _, v := range entries {
-		value := deepGet(*v, key)
-		if value != nil {
-			items := strings.Split(value.(string), sep)
-			for _, item := range items {
-				groups[item] = append(groups[item], v)
-			}
-
-		}
-	}
-	return groups
-}
-
-// groupBy groups a list of *RuntimeContainers by the path property key
-func groupBy(entries []*RuntimeContainer, key string) map[string][]*RuntimeContainer {
-	groups := make(map[string][]*RuntimeContainer)
-	for _, v := range entries {
-		value := deepGet(*v, key)
-		if value != nil {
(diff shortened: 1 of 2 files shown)


md5:
Nice work! :+1: 

Sorry for the breakage, but it's good to see both the `where*` functions and the `group*` functions generalized.

jwilder:
Thanks @webner @md5 !

template.go
```diff
@@ -29,52 +29,84 @@ func exists(path string) (bool, error) {
...
-		}
+	kind := entriesVal.Kind()
+
+	if kind == reflect.Ptr {
+		entriesVal = reflect.Indirect(entriesVal)
+		kind = entriesVal.Kind()
 	}
-	return groups
+
+	switch entriesVal.Kind() {
```
md5: Sorry I didn't notice this until after the merge, but shouldn't this be `switch kind {`?

Perhaps this can be the fix that gets put into #100, since I think it will have the same issue of returning an error when given a pointer to a slice or an array.

template.go
```diff
@@ -29,52 +29,84 @@ func exists(path string) (bool, error) {
...
+	case reflect.Array, reflect.Slice:
+		break
+	default:
+		return entriesVal, fmt.Errorf("Must pass an array or slice to '%v'; received %v; kind %v", funcName, entries, kind)
```
md5: This is also weird. Shouldn't it return `nil` and the error, not `entriesVal`?

template.go
webner: You are right `kind` is better and reduces one call to `Kind()`. (It wont change the result because entriesVal is also updated when the pointer is dereferenced)

I think the issue #99 is already fixed with this PR and #100 is not needed.

template.go
webner: Yes, good catch. I also will change this to nil.

webner:
I opened a new PR with the fixes #101

md5:
:+1:

PR #98, opened 2015-07-13
Added nginx-ssl-include.tmpl template optimized for ownCloud and Seafile.

ypid:
Example usage: https://github.com/ypid/docker-makefile/blob/master/Makefile

### DIFF
--- a/templates/nginx-ssl-include.tmpl
+++ b/templates/nginx-ssl-include.tmpl
@@ -1,0 +1,120 @@
+# See https://github.com/ypid/docker-makefile/blob/master/Makefile as usage example.
+{{ $proto := "http" }}
+    listen 443;
+    listen [::]:443;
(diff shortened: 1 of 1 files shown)


jwilder:
Not sure if you meant to make a change for `nginx-proxy` or just add this as an example template.  This is too specific to include in `nginx-proxy` though.

ypid:
I wanted to add another, more suphisticated example. Should I submit it to https://github.com/jwilder/nginx-proxy instead?

jwilder:
Ok.  Can you name the template after the package it's for and add it to the `templates` dir?  See: https://github.com/jwilder/docker-gen/tree/master/templates for other examples.

ypid:
OK.

ypid:
Any news?

PR #104, opened 2015-07-23
skip-blank-lines option

### LINKED ISSUE #16: Blank Lines In Generated Configuration
While this issue is cosmetic in nature I was wondering if there's any way it can be fixed/prevented as readability is severely impaired with many hosts.

An example template I use is here: https://github.com/maxexcloo/Docker/blob/master/apps/haproxy-config/haproxy.tmpl

webner:
fixes #16

### DIFF
--- a/template.go
+++ b/template.go
@@ -382,7 +382,7 @@ func generateFile(config Config, containers Context) bool {
 
 	contents := executeTemplate(config.Template, filteredContainers)
 
-	if config.SkipBlankLines {
+	if !config.KeepBlankLines {
 		contents = removeBlankLines(contents)
 	}
 
--- a/template.go
+++ b/template.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"bufio"
 	"bytes"
 	"crypto/sha1"
 	"encoding/json"
@@ -383,7 +382,9 @@ func generateFile(config Config, containers Context) bool {
 	contents := executeTemplate(config.Template, filteredContainers)
 
 	if !config.KeepBlankLines {
-		contents = removeBlankLines(contents)
+		buf := new(bytes.Buffer)
+		removeBlankLines(bytes.NewReader(contents), buf)
+		contents = buf.Bytes()
 	}
 
 	if config.Dest != "" {
@@ -433,22 +434,10 @@ func executeTemplate(templatePath string, containers Context) []byte {
 		log.Fatalf("unable to parse template: %s", err)
 	}
 
-	var buf bytes.Buffer
-	err = tmpl.ExecuteTemplate(&buf, filepath.Base(templatePath), &containers)
+	buf := new(bytes.Buffer)
+	err = tmpl.ExecuteTemplate(buf, filepath.Base(templatePath), &containers)
 	if err != nil {
 		log.Fatalf("template error: %s\n", err)
 	}
 	return buf.Bytes()
-}
-
-func removeBlankLines(buf []byte) []byte {
-	filtered := new(bytes.Buffer)
-	scanner := bufio.NewScanner(bytes.NewReader(buf))
-	for scanner.Scan() {
-		line := scanner.Text()
-		if !isBlank(line) {
-			fmt.Fprintln(filtered, line)
-		}
-	}
-	return filtered.Bytes()
 }
--- a/template.go
+++ b/template.go
@@ -410,7 +410,7 @@ func generateFile(config Config, containers Context) bool {
 		scanner := bufio.NewScanner(bufio.NewReader(&buf))
 		for scanner.Scan() {
 			line := scanner.Text()
-			if len(line) > 0 {
+			if !isBlank(line) {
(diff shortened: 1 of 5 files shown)


md5:
I think this should probably also omit lines that are all whitespace, not just empty lines.

template.go
```diff
@@ -398,11 +399,24 @@ func generateFile(config Config, containers Context) bool {
 	}
 
 	var buf bytes.Buffer
-	multiwriter := io.MultiWriter(dest, &buf)
-	err = tmpl.ExecuteTemplate(multiwriter, filepath.Base(templatePath), &filteredContainers)
+	bw := bufio.NewWriter(&buf)
```
jwilder: This seems to have broken the code that compares if the generated file is different from the existing one.  

Running `docker-gen -skip-blank-lines ./templates/nginx.tmpl /tmp/nginx` always generates a new `/tmp/nginx` even if the contents have not changed.    The issue is that `bytes.Buffer` clears it's buffer after `WriteTo` is called on L418.  So the comparison on L437 always fails.

See: http://play.golang.org/p/tcz4g_VHZ_

jwilder:
@webner One regression, but this will be nice feature to add.

thaJeztah:
Are there downsides to this? Otherwise I suggest reversing the flag, and make this the default (I.e. Rename the flag to `--keep-blank-lines`) :+1:

md5:
The downside is that it could be a breaking change if anyone has managed to
use docker-gen in a setup that has significant whitespace. Doing so would
be a huge pain in the ass, but I can't rule out the possibility that
someone has done so.

thaJeztah:
In that case, the `--keep-blank-lines` option provides a way out. Sure, it should be mentioned in the release note as a "potentially breaking change"

md5:
@thaJeztah I hadn't thought about the fact that `docker-gen` is actually versioned (and with a 0.x version no less), so perhaps that would be reasonable.

webner:
I like the idea of making it the default and renaming the option to `--keep-blank-lines`
@jwilder what do you think?

jwilder:
@webner sounds good

md5:
Latest diff LGTM

md5:
I just realized that the updated code will not preserve the original line endings in the template file. Could that be a problem?

webner:
I will change it so that it will preserve the line endings

md5:
That looks fine as long as nobody uses super old-school `\r` only Mac files.

thaJeztah:
> as long as nobody uses super old-school \r-only Mac files.

If they are, they are just doing it wrong, so :+1:

webner:
If you use `\r` line endings at least it won't break your output and leave it as-is. ;)

md5:
:+

md5:
I just found out that a change has been committed for `text/template` to allow trimming whitespace: https://github.com/golang/go/commit/e6ee26a03b79d0e8b658463bdb29349ca68e1460

I'd still like to see this change here merged, but I'm glad to see that better whitespace control is coming to a future Go version.

jwilder:
I can't seem to get this to generate a file to disk anymore.  
```
$ docker-gen ./templates/nginx.tmpl /tmp/nginx
2015/09/12 15:57:12 Contents of /tmp/nginx did not change. Skipping notification ''
```
But I don't have a `/tmp/nginx` file on disk before or after I run this.

webner:
are you sure you tested the latest commit?

```
$ git status
On branch skip-blank-lines
Your branch is up-to-date with 'webner/skip-blank-lines'.
nothing to commit, working directory clean

$ git rev-list HEAD --max-count=1
480ed4c970db560211833c24e5199bcad786eb34

$ make
Building docker-gen
# github.com/jwilder/docker-gen
link: warning: option -X main.buildVersion 0.4.0-12-g480ed4c may not work in future releases; use -X main.buildVersion=0.4.0-12-g480ed4c

$ ./docker-gen ./templates/nginx.tmpl /tmp/nginx
2015/09/15 11:24:09 Generated '/tmp/nginx' from 0 containers

$ ./docker-gen ./templates/nginx.tmpl /tmp/nginx
2015/09/15 11:24:13 Contents of /tmp/nginx did not change. Skipping notification ''

$ rm /tmp/nginx

$ ./docker-gen ./templates/nginx.tmpl /tmp/nginx
2015/09/15 11:24:21 Generated '/tmp/nginx' from 0 containers

$ cat /tmp/nginx
server {
        listen 80 default_server;
        server_name _; # This is just an invalid value which will never trigger on a real hostname.
        error_log /proc/self/fd/2;
        access_log /proc/self/fd/1;
        return 503;
}
```

aderouineau:
I sure welcome this pull request!

@jwilder Could it be quickly merged, and a new docker-gen image created? Thanks!

jwilder:
Thanks @webner!

aderouineau:
@jwilder Time for 0.5?

PR #108, opened 2015-08-05
added when template function

webner:
The `when` function can be used like the conditional (ternary) operator in javascript/c++/c#. 
` condition ? trueValue : falseValue `

Examples:
```
{{ $color := when $useDarkTheme "blue" "white" }}
{{ $defaultServer := when (eq $host $defaultHost) "default" "" }}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -233,6 +233,7 @@ For example, this is a JSON version of an emitted RuntimeContainer struct:
 * *`whereNotExist $items $fieldPath`*: Like `where`, but returns only items where `$fieldPath` does not exist (is nil).
 * *`whereAny $items $fieldPath $sep $values`*: Like `where`, but the string value specified by `$fieldPath` is first split by `$sep` into a list of strings. The comparison value is a string slice with possible matches. Returns items which OR intersect these values.
 * *`whereAll $items $fieldPath $sep $values`*: Like `whereAny`, except all `$values` must exist in the `$fieldPath`.
+* *`when $condition $trueValue $falseValue`*: Returns the `$trueValue` when the `$condition` is `true` and the `$falseValue` otherwise
 
 ===
 
--- a/template.go
+++ b/template.go
@@ -327,6 +327,15 @@ func trimSuffix(suffix, s string) string {
 	return strings.TrimSuffix(s, suffix)
 }
 
+// when returns the trueValue when the condition is true and the falseValue otherwise
+func when(condition bool, trueValue, falseValue interface{}) interface{} {
+	if condition {
+		return trueValue
+	} else {
+		return falseValue
+	}
+}
+
 func newTemplate(name string) *template.Template {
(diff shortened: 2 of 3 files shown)


md5:
This LGTM. It might be nice to extend this at some point to support non-boolean truthy values the same way that `if` does, but that would be a backwards compatible change if it's added later.

webner:
I would like to add to support non-boolean truthy. But the only way I could think of was to copy the whole `isTrue` function from `text/template/exec.go` to make it consistent with the other build-in functions. I'm not sure if it is worth it. The workaround is to use `not` to convert it to a boolean value.
e.x. ` {{ $isTrue := not (not $anyValue) }}`

```
// isTrue reports whether the value is 'true', in the sense of not the zero of its type,
// and whether the value has a meaningful truth value.
func isTrue(val reflect.Value) (truth, ok bool) {
	if !val.IsValid() {
		// Something like var x interface{}, never set. It's a form of nil.
		return false, true
	}
	switch val.Kind() {
	case reflect.Array, reflect.Map, reflect.Slice, reflect.String:
		truth = val.Len() > 0
	case reflect.Bool:
		truth = val.Bool()
	case reflect.Complex64, reflect.Complex128:
		truth = val.Complex() != 0
	case reflect.Chan, reflect.Func, reflect.Ptr, reflect.Interface:
		truth = !val.IsNil()
	case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
		truth = val.Int() != 0
	case reflect.Float32, reflect.Float64:
		truth = val.Float() != 0
	case reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64, reflect.Uintptr:
		truth = val.Uint() != 0
	case reflect.Struct:
		truth = true // Struct values are always true.
	default:
		return
	}
	return truth, true
}
```

md5:
@webner Probably not worth it since it can be easily added later if someone really wants it.

md5:
I opened an issue with Go itself it see if they will expose `isTrue`: https://github.com/golang/go/issues/12033

n-rodriguez:
why not using ParseBool function from strconv package?

my 2 cents

md5:
`ParseBool` requires a `string` as input while `isTrue` can take any value (e.g. treating the number `0` as false).

n-rodriguez:
Thanks for you answer :)

jwilder:
Thanks @webner @md5!

md5:
:metal: 

BTW, `IsTrue` did end up getting exposed (cf. https://github.com/golang/go/commit/a326c3e1ad3713c3e1c3373a45c6907e10fb1579), so it would be cool to enhance this once Go 1.6 is released.

md5:
Go 1.6 was released this week, including the `IsTrue` function in `text/template`, so it would now be possible to take any `interface {}` as the `condition` argument.

PR #110, opened 2015-08-16
Export ParseBool function from strconv package to template

n-rodriguez:
btw: the first lines of Go in my life :)

### DIFF
--- a/template.go
+++ b/template.go
@@ -13,6 +13,7 @@ import (
 	"os"
 	"path/filepath"
 	"reflect"
+	"strconv"
 	"strings"
 	"syscall"
 	"text/template"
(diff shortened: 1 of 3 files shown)


template.go
```diff
@@ -346,6 +351,7 @@ func newTemplate(name string) *template.Template {
 		"keys":          keys,
 		"last":          arrayLast,
 		"replace":       strings.Replace,
+		"parseBool":     parseBool,
```
md5: You may just want to do `"parseBool": strconv.ParseBool` here since the function signature and return type isn't changing.

md5:
Nice work!

jwilder:
Thanks @n-rodriguez!

n-rodriguez:
You're welcome! This is a very nice tool to have :)

PR #113, opened 2015-08-20
fix an incorrect docker-gen parameter in example, other minor fixes and 1 proposed change to example

RichLewis007:
- Needed to fix the incorrect --only-published to -only-published 
- Proposed change: I think it's more likely to work for more people if change -only-published to -only-exposed
- added :ro to the sock, to match your other examples in your other project pages.


jwilder:
Thanks @RichLewis007!

PR #116, opened 2015-09-09
Add support for docker 1.8 Mounts

cybercode:
Also silenced warning about ldflags in makefile.

### DIFF
--- a/docker-gen.go
+++ b/docker-gen.go
@@ -58,6 +58,13 @@ type Volume struct {
 	ReadWrite bool
 }
 
+type Mount struct {
+	Source      string
+	Destination string
+	Mode        string
+	RW          bool
+}
+
 type RuntimeContainer struct {
 	ID        string
 	Addresses []Address
@@ -67,6 +74,7 @@ type RuntimeContainer struct {
 	Image     DockerImage
 	Env       map[string]string
 	Volumes   map[string]Volume
+	Mounts    map[string]Mount
 	Node      SwarmNode
 	Labels    map[string]string
 	IP        string
--- a/GLOCKFILE
+++ b/GLOCKFILE
@@ -1,4 +1,4 @@
 github.com/BurntSushi/toml f87ce853111478914f0bcffa34d43a93643e6eda
 github.com/Sirupsen/logrus 6ebb4e7b3c24b9fef150d7693e728cb1ebadf1f5
-github.com/docker/docker 95dcb2748442fc1f59e8c2e010679e0aea3cef1c
-github.com/fsouza/go-dockerclient 1e60afa8d58b91a0eca0a732927c8ecba696493f
+github.com/docker/docker b66e5ef20824ecef46d657833db409a035b65306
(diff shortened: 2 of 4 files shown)


docker-gen.go
```diff
@@ -58,6 +58,13 @@ type Volume struct {
 	ReadWrite bool
 }
 
+type Mount struct {
+	Source      string
+	Destination string
+	Mode        string
+	RW          bool
```
jwilder: Can you rename this to `ReadWrite` to be consistent w/ `Volume`?

cybercode: On Wed, Sep 09, 2015 at 11:18:56AM -0700, Jason Wilder wrote:
>    In docker-gen.go:
>
>  > @@ -58,6 +58,13 @@ type Volume struct {
>  >    ReadWrite bool
>  >  }
>  >
>  > +type Mount struct {
>  > +  Source      string
>  > +  Destination string
>  > +  Mode        string
>  > +  RW          bool
>
>    Can you rename this to ReadWrite to be consistent w/ Volume?

Since 'ReadWrite' is synthesized from 'VolumesRW' in APIs < 1.20, doesn't it
sense to maintain consistency w/the API and dockerclient in this case, where
the 'Mount' struct is a verbatim copy from go-dockerclient
(https://github.com/fsouza/go-dockerclient/blob/master/container.go#L224.)

Granted i have deviated from the underlying library by turning the Mounts
array into a map, but i feel this makes it much more useful in the template
use-case (FWIW, i feel the API implementation of the mounts as an array
rather than a map is a regression from the previous version).

rick

jwilder: docker-gen exposes it's own API that I'd like to keep consistent.  It's not the same as go-dockerclient even though it uses it.

cybercode: On Sat, Sep 12, 2015 at 02:45:21PM -0700, Jason Wilder wrote:
>    In docker-gen.go:
>
>  > @@ -58,6 +58,13 @@ type Volume struct {
>  >    ReadWrite bool
>  >  }
>  >
>  > +type Mount struct {
>  > +  Source      string
>  > +  Destination string
>  > +  Mode        string
>  > +  RW          bool
>
>    docker-gen exposes it's own API that I'd like to keep consistent. It's not
>    the same as go-dockerclient even though it uses it.
>

If that is the case, since Volumes and Mounts are mutually exclusive, how
about not adding a new `Mounts` structure at all and filling in the `Volumes`
structure w/ the mounts info:

	for _, v := range container.Mounts {
		runtimeContainer.Volumes[v.Destination] = Volume{
			HostPath:  v.Source,
			Path:      v.Destination,
			ReadWrite: v.RW,
		}
	}

That way the docker-gen api doesn't change at all, regardless of the
go-dockerclient version. Also, it means the only change to the source is the
addition of the above. No changes to data structures, etc.

rick

cybercode:
Sending new request which doesn't add new map (iupdates Volumes)

PR #118, opened 2015-09-11
Adding JSON and YAML support for the configuration file

asdine:
In some workflows, the configuration file is generated dynamically
It would be nice if more formats where supported to provide much more options to generate it.

It is also nicer for users to use their favorite file format


jwilder:
Interesting idea.  Can you elaborate on your use case of generating the config?  What are you generating it from?

asdine:
The file can be generated by a back office who allows to enable/disable a service and its containers, or to "move" some from one node to another. The idea is to let the back office generate the configuration file and pass it to Ansible or Salt who then delivers it on the selected node.

Supporting multiple configuration formats allows anyone to create a configuration file from anywhere.

jwilder:
@asdine Thanks for the PR.  I'm closing this as I don't think it makes sense to have multiple config formats.  TOML is a text based format and pretty easy to generate.  Adding YAML and JSON would add more complexity and dependencies.

PR #119, opened 2015-09-14
Add support for docker 1.8 Mounts

cybercode:
This patch uses the current `Volumes` map instead of creating a new `Mounts` map. This should make the underlying API changes transparent to docker-gen users.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 FROM debian:wheezy
 MAINTAINER Jason Wilder <jason@influxdb.com>
 
-ENV VERSION 0.5.0
+ENV VERSION 0.5.1
 ENV DOWNLOAD_URL https://github.com/jwilder/docker-gen/releases/download/$VERSION/docker-gen-linux-amd64-$VERSION.tar.gz
 ENV DOCKER_HOST unix:///tmp/docker.sock
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,21 +1,21 @@
 FROM debian:wheezy
 MAINTAINER Jason Wilder <jason@influxdb.com>
 
-ENV VERSION 0.4.1
+ENV VERSION 0.5.0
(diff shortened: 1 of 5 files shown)


jwilder:
I'm not sure if it's better to reuse the `Volumes` key, create a new `Mounts` key or do both and deprecate `Volumes`.  Seems like `Volumes` has been removed in favor of `Mounts` in new docker versions but I couldn't find any official documentation confirming that.

I'm thinking it would be better to re-use both `Volumes` and add a new `Mounts` key to keep some backwards compatibility but also keep the terminology consistent w/ what you see in `docker inspect`.  In  future release, `Volumes` could be removed.  Thoughts?

cybercode:
On Tue, Sep 22, 2015 at 09:09:36AM -0700, Jason Wilder wrote:
>    I'm not sure if it's better to reuse the Volumes key, create a new Mounts
>    key or do both and deprecate Volumes. Seems like Volumes has been removed
>    in favor of Mounts in new docker versions but I couldn't find any official
>    documentation confirming that.

The docker people don't seem to believe in semantic versioning, Change notes
or deprecation. The only way to figure out version to version changes is the
API documentation -- which no longer has Volumes in the runtime data
structure, so yes, it's been removed. However, note that Volumes still as
part of the POST api. I'm guessing their thinking is to separate persistent
volumes from mounted host directories.

>    I'm thinking it would be better to re-use both Volumes and add a new
>    Mounts key to keep some backwards compatibility but also keep the
>    terminology consistent w/ what you see in docker inspect. In future
>    release, Volumes could be removed. Thoughts?

I think the best way to handle the change is in three steps (esp. since
many -- including me -- seem to need a 1.8 compatible version of nginx-proxy):

  1. Reuse the Volumes map, allowing transparent updating to docker 1.8.x
  2. Add the Mounts, using the conventions of underlying map: Source,
     Destination, RW) and deprecating Volumes.
  3. Remove Volumes

If you want to combine 1 & 2, it's ok w/ me (i will push another PR adding
back the Mounts.) Personally, i would do them in separate steps pushing
the current Volumes PR first.

rick

jwilder:
Fixed via #146.

PR #120, opened 2015-09-23
strip www from host names

RichLewis007:
If a host name begins with "www." strip it off and redirect to the naked host name.

I don't know if you want to add this to your code, but I find it helpful as I have both www.example.com and example.com coming in to be handled by nginx and sent to containers. But I prefer all users get redirected to see just the example.com URL in their browsers.

### DIFF
--- a/templates/nginx.tmpl
+++ b/templates/nginx.tmpl
@@ -40,6 +40,12 @@ upstream {{ $host }} {
 {{ end }}
 }
 
+# strip www from any host names which begin with it
+server {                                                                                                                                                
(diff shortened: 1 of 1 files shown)


RichLewis007:
If a host name begins with "www." strip it off and redirect to the naked host name.

I don't know if you want to add this to your code, but I find it helpful as I have both www.example.com and example.com coming in to be handled by nginx and sent to containers. But I prefer all users get redirected to see just example.com in their browsers.

jwilder:
Thanks.  That file is just an example starter template.

PR #122, opened 2015-10-07
added ipv6 support

lukas2511:
These changes add IPv6 support to docker-gen.

IPv6 addresses of docker containers are available as `IP6_LinkLocal` and `IP6_Global` next to the old `IP` string.

I had to update the used docker client library to a newer version with support for `container.NetworkSettings.LinkLocalIPv6Address` and `container.NetworkSettings.GlobalIPv6Address`

### DIFF
--- a/docker-gen.go
+++ b/docker-gen.go
@@ -47,8 +47,8 @@ type Event struct {
 
 type Address struct {
 	IP            string
-	IP6_LinkLocal string
-	IP6_Global    string
+	IP6LinkLocal  string
+	IP6Global     string
 	Port          string
 	HostPort      string
 	Proto         string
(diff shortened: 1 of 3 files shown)


docker-gen.go
```diff
@@ -46,11 +46,13 @@ type Event struct {
 }
 
 type Address struct {
-	IP       string
-	Port     string
-	HostPort string
-	Proto    string
-	HostIP   string
+	IP            string
+	IP6_LinkLocal string
```
jwilder: Could you remove the underscores `_`?  I try to keep the code consistent with: https://golang.org/doc/effective_go.html#mixed-caps

lukas2511:
As you wished I removed the underscores from the name of the IPv6 fields

jwilder:
Thanks @lukas2511!

PR #123, opened 2015-10-08
Added trim to available methods in template

mariusGundersen:
it simply removes whitespace from the given string.


jwilder:
Thanks @mariusGundersen!

md5:
I ran into a problem with this in https://github.com/jwilder/nginx-proxy/commit/f39459680659971e0115ce7a0432b2a47ed5cb11

Since `strings.TrimSpace` strictly takes a `string`, it barfs if it gets passed an `interface{}`, even if it's actually a `string`.

I think it would be worthwhile to have some sort of `stringToStringHelper` function to use internally in `docker-gen` to take functions that are `func(string) string` and turn them into `func(interface {}) string` using reflection to check if the argument is actually a `string` or a `*string`. I could also see having other helpers like `stringStringToStringHelper` for the `func(string,string) string` case (and possibly others).

jwilder:
Maybe all the template functions should just working with `interface{}` and return the original value if it's not a string?

md5:
I think they should probably error out when they aren't given what they expect, but I think in general it's good for the functions to be flexible in the concrete types they accept as long as the reflected types are usable.

jwilder:
Agreed.  Looks like hugo does that for its functions: https://github.com/spf13/hugo/blob/master/tpl/template_funcs.go

md5:
[`spf13/cast`](https://github.com/spf13/cast) looks quite useful :+1:

Issue #124, opened 2015-10-09
Separate Container Install: nginx.tmpl:52: function "trim" not defined


Boran:
Just tried the cut-and-paste example for 'Separate Container Install', nginx comes up fine, but nginx-gen gives
```
2015/10/09 06:40:41 unable to parse template: template: nginx.tmpl:52: function "trim" not defined
````
There is a trim() in https://github.com/jwilder/docker-gen/blob/master/template.go#L332, how come it does not find it?
Perhaps the image on dockerhub is behind this repo (and its readme)?
So I cloned this repo and build the image locally, but the same result.

Suggestions?

md5:
The `trim` function is not in the 0.4.1 release of `docker-gen` that the image is using. The `Dockerfile` needs to be updated to pull in the latest 0.4.2 release that @jwilder pushed out this morning.

Boran:
I updated Dockerfile and build locally and yes it built and runs fine.
Thanks.

md5:
@Boran If you want to make a PR, I'm sure @jwilder would appreciate it.

jwilder:
:+1:

Boran:
Well there you go. Hopefully I got that complex patch right :-)

PR #125, opened 2015-10-09
VERSION 0.4.2 in Dockerfile

Boran:
As per issue #124, the version number was not incremented in the Dockerfile for that release


jwilder:
Thanks @Boran!

PR #126, opened 2015-10-17
Add splitN

md5:


md5:
Thanks @jwilder. If you can cut a new release, I'll see about the combining the path support for `jwilder/nginx-proxy` into a single environment variable. :+1:

jwilder:
@md5 Done.

PR #128, opened 2015-10-19
Add server info

tedkulp:
Uses the info and version commands to get various info about the docker server.

My use-case for this is forwarding a lot of information to logstash from various clusters and wanting to be able to track down the server that the log entry is coming from quickly. Since fleet can deploy containers to different servers based on rules and availability, getting the hostname that the container lives on is vital.

I've also updated the structs in the README to match what's currently in the code.

Let me know if you have any questions/changes/etc.

Thanks!
Ted

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,17 +135,20 @@ Within the templates, the object emitted by docker-gen will be a structure consi
 
 ```go
 type RuntimeContainer struct {
-    ID        string
-    Addresses []Address
-    Gateway   string
-    Name      string
-    Hostname  string
-    Image     DockerImage
-    Env       map[string]string
-    Volumes   map[string]Volume
-    Node      SwarmNode
-    Labels    map[string]string
-    IP        string
+	ID           string
+	Addresses    []Address
+	Gateway      string
+	Name         string
+	Hostname     string
+	Image        DockerImage
+	Env          map[string]string
+	Volumes      map[string]Volume
+	Node         SwarmNode
+	Labels       map[string]string
+	Server       ServerInfo
+	IP           string
+	IP6LinkLocal string
+	IP6Global    string
 }
 
 type Address struct {
@@ -172,6 +175,21 @@ type SwarmNode struct {
     ID      string
     Name    string
     Address Address
+}
+
+type ServerInfo struct {
+	Name          string
+	NumContainers int
+	NumImages     int
+	DockerInfo    DockerInfo
+}
+
+type DockerInfo struct {
+	Version         string
+	ApiVersion      string
+	GoVersion       string
+	OperatingSystem string
+	Architecture    string
 }
 ```
 
--- a/docker-gen.go
+++ b/docker-gen.go
@@ -72,6 +72,7 @@ type RuntimeContainer struct {
 	Volumes      map[string]Volume
 	Node         SwarmNode
 	Labels       map[string]string
+	Server       ServerInfo
 	IP           string
 	IP6LinkLocal string
 	IP6Global    string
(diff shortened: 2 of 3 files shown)


md5:
This LGTM. This diff is easier to review: https://github.com/jwilder/docker-gen/pull/128/files?w=1

One question I have is why the check for `apiVersion != nil` is inside the `apiInfo != nil` check. Seems like they could be independent.

tedkulp:
Only because it's nested inside. I could split them out into two different structs next to each other if you think it makes more sense. I was trying to not make the root struct too messy.

md5:
​Sounds good. Let's see what @jwilder thinks.

README.md
```diff
@@ -135,17 +135,20 @@ Within the templates, the object emitted by docker-gen will be a structure consi
...
+	Addresses    []Address
+	Gateway      string
+	Name         string
+	Hostname     string
+	Image        DockerImage
+	Env          map[string]string
+	Volumes      map[string]Volume
+	Node         SwarmNode
+	Labels       map[string]string
+	Server       ServerInfo
```
jwilder: Maybe name this `Docker` instead?  `Server` is kind of nebulous.

README.md
```diff
@@ -173,6 +176,21 @@ type SwarmNode struct {
     Name    string
     Address Address
 }
+
+type ServerInfo struct {
+	Name          string
+	NumContainers int
+	NumImages     int
+	DockerInfo    DockerInfo
```
jwilder: If `ServerInfo` is renamed to `Docker`, field could renamed to just `Info`.  So the the template looks like `Docker.Info`.

docker_client.go
```diff
@@ -155,6 +166,20 @@ func getContainers(client *docker.Client) ([]*RuntimeContainer, error) {
 				ReadWrite: container.VolumesRW[k],
 			}
 		}
+		if apiInfo != nil {
```
jwilder: I think this could be done once before getting the containers.  All the containers can just have point to the same reference since it's all the same.

docker-gen.go
```diff
@@ -72,6 +72,7 @@ type RuntimeContainer struct {
 	Volumes      map[string]Volume
 	Node         SwarmNode
 	Labels       map[string]string
+	Server       ServerInfo
```
jwilder: Seems like this should go on the `Context` and not on each `RuntimeContainer`.  That way you could access it directly without a reference to a container.  

See: https://github.com/jwilder/docker-gen/blob/master/docker-gen.go#L131

tedkulp: I think that's a great idea. Can you get to the Context from inside a template, though? It looks like it only sends an array of containers to the generateFile function.

docker_client.go
tedkulp: Agreed. I'll fix that once we sort out the Context thing.

docker-gen.go
md5: It seems like putting it on `Context` would be quite a bit of effort and wouldn't work that well.

Assuming there would be a `func (c *Context) Docker() *Docker` function, that function would have to call `getEndpoint()` and `NewDockerClient()` to create a Docker client each time it's called since it has no access to an existing `docker.Client`. I also can't see how `docker-gen` would be able to avoid hitting the Docker server every time the function is called.

I suppose you could stash a reference to the `docker.Client` inside of each `RuntimeContainer`, but that seems like a terrible hack.

I could see making it work if `Context` were turned into a `struct`, but then it would not be possible to call `range` on it and iterate over a list of `RuntimeContainer` instances without changing every existing `docker-gen` template.

tedkulp: Just a poke to keep this change alive. I have been using it in production here for the past few days without any changes and it's been working well. I do agree that copying the same info over to every RuntimeContainer object is probably not efficient, but as far as being practical, it's been good.

jwilder:
Fixed via #146.

PR #137, opened 2015-11-27
Add rescan of containers after Docker API didn't reply on ping

### LINKED ISSUE #136: Recheck containers after sleep on 'Unable to ping docker daemon' error
I've had a weird issue with jwilder/nginx-proxy that didn't work properly up after a system restart. The proxy container had restart=always, no dependencies and was probably started first. Somehow docker 1.9.1 API returns '404 page not found' error at the very beginning. Probably the container was up before API become available.
```
dockergen.1 | 2015/11/23 19:38:48 error listing containers: API error (404): 404 page not found
dockergen.1 | 2015/11/23 19:38:48 Unable to ping docker daemon: API error (404): 404 page not found
dockergen.1 | 2015/11/23 19:38:58 Watching docker events
```

Now, when docker-gen receives an error on a ping request it just sleeps for 10 seconds. After that no check is being made for containers that happen to come up while it was sleeping. So it misses all the containers that have restart=always on boot. 

And this kinda sucks :) Have to add a dependency to nginx-proxy so it won't be started first or restart the proxy container manually after boot.

dadittoz:
This fixes #136

### DIFF
--- a/docker-gen.go
+++ b/docker-gen.go
@@ -305,6 +305,8 @@ func generateFromEvents(client *docker.Client, configs ConfigFile) {
 	wg.Add(1)
 	defer wg.Done()
 
+	shouldRescanContainters := false
+
 	for {
 		if client == nil {
 			var err error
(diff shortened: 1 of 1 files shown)


jwilder:
Sounds like this might have been a race in docker.

Issue #75, opened 2015-03-16
forward stdout/stderr of -notify command


stefanfoulis:
forward stdout/stderr of ``-notify`` command
This can greatly help when things go wrong. E.g when your command is ``-notify "nginx -s reload"`` and for some reason there is a syntax error in the generated config. Then you'd get the error from nginx directly and would not have to dive into the container to debug.

ripperdoc:
+1 on this!

PR #142, opened 2015-12-16
Add notify output

bcicen:
Adds option to forward any output from the notify command to the logstream of docker-gen.  Resolves https://github.com/jwilder/docker-gen/issues/75


jwilder:
Thanks @bcicen!

PR #143, opened 2015-12-17
Add whereLabel* functions

md5:
This PR adds a few functions from the `whereLabel*` family:

* `whereLabelExists`
* `whereLabelDoesNotExist`
* `whereLabelValueMatches`

This should help with situations like those mentioned in https://github.com/jwilder/docker-gen/pull/82#issuecomment-158323192

Right now, these functions are strictly for `[]*RuntimeContainer`, but they could be adjusted to work in other situations if need be. I could also see adding a `hasLabel` function that takes `*RuntimeContainer` or exposing the `regexp.MatchString` function used by `whereLabelValueMatches`, but I wanted to keep this PR simple and focused.


md5:
An example use case (as I mentioned in https://github.com/jwilder/docker-gen/pull/82#issuecomment-165362470) would be:

```
whereLabelValueMatches . "com.docker.compose.service" "^web$"
```

jwilder:
Thanks @md5!

md5:
:+1:

PR #144, opened 2015-12-18
Add support for NetworkSettings.Networks

dustinblackman:
## Overview

Latest version of Docker and the go docker client adds `NetworkSettings.Networks`, which contains information for Docker's [multi-host networking](https://docs.docker.com/engine/userguide/networking/get-started-overlay/). This adds support for applications such as your [nginx-proxy](https://github.com/jwilder/nginx-proxy) to route to containers internally across Docker's networks without having to expose ports.

I currently have this running in a production cluster and looks good!

PR #145, opened 2015-12-23
Update documentation on available structs

punkstar:
* Document the Network struct introduced by #144
* Document the IPv6 support introduced by #122


jwilder:
Thanks!

PR #153, opened 2016-01-15
Deal with networks (supports new overlay network)

baptistedonaux:
Initial template works with current model network.

If you use the new overlay network (ex. docker-compose --x-networking up), the golang structure RuntimeContainer -> []Address doesn't contain IP (empty string).

Use []Network array resolved the problem.

### DIFF
--- a/templates/nginx.tmpl
+++ b/templates/nginx.tmpl
@@ -12,20 +12,22 @@ upstream {{ $host }} {
 {{ range $index, $value := $containers }}
 
 	{{ $addrLen := len $value.Addresses }}
+	{{ $network := index $value.Networks 0 }}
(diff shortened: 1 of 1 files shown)


jwilder:
This file is just an example.  Were you intending to fix this issue w/ `nginx-proxy`?

baptistedonaux:
I send PR on ```jwilder/nginx-proxy```.

But I think that fix template example right now will limit the feedback, because the new model network will be activate systematically in the ```docker-compose``` next release. For docker-compose 1.6, the containers IP will be set in []Network and not more in []Address.

jwilder:
Thanks @baptistedonaux!

Issue #103, opened 2015-07-22
Notify throttling


ybogdanov:
Hi,
It would be nice to have some kind of configurable notification delay. For example, if I run 10 containers in a row within a second, it generates dozens of events and each will cause SIGHUP notify or equivalent. 

With 5 seconds delay, docker-gen may wait and then notify only once.

I can contribute this functionality after you confirm the feature is needed and will be accepted.

jwilder:
Is this causing any specific issues in your case?  Sound similar to #42.

ybogdanov:
@jwilder yes, it's the same. In case of using long running connections, it might be expensive to reload nginx to frequently, because it keeps old worker processes until all connections are closed. It consumes memory and under load may cause OOM kills. Also, depending on the design of upstream servers, it could be expensive to simply cut all connections and let all connections (or part, in case of using multiple balancers) to reconnect.

I can imagine tools that is even more sensitive to reloads/restarts than nginx. So I'm sure we need some king of delay to avoid unnecessary reloads or restarts.

jwilder:
Fixed via #156

Issue #57, opened 2015-02-17
Race condition in -watch


mhsmith:
We are using docker-gen in this dynamic DNS server: https://github.com/ncarlier/dockerfiles/tree/master/dnsmasq. This requires that the DNS configuration is updated when docker-gen is started, and whenever containers are started and stopped thereafter. Unfortunately, there is a gap between the initial update and the point where docker-gen starts watching for events, as shown in this log:

    2015-02-11 18:45:42,946 INFO spawned: 'docker-gen' with pid 9
    2015-02-11 18:45:43,001 INFO spawned: 'dnsmasq' with pid 10
    2015/02/11 18:45:43 Generated '/etc/dnsmasq.conf' from 4 containers
    2015/02/11 18:45:43 Running 'supervisorctl restart dnsmasq'
    [...]
    2015/02/11 18:45:45 Watching docker events

Any containers started during that 2-second window will not appear in DNS until the next docker event, which may be an indefinite period of time.

Solution: docker-gen should start watching for events before its initial update.

jwilder:
#147 might help here.

mitar:
How would HUP help here?

gigablah:
It manually triggers docker-gen to do a re-scan of containers.

jwilder:
Fixed via #156

Issue #43, opened 2014-12-02
delay on notify


jgallen23:
my server takes a little bit of time to start up, so I'd like to have a delay before notifying nginx and adding to the proxy.  This sort of ties into #42, as I'd only want a single event if multiple were fired within the delay interval

RoboMod:
+1

jgallen23:
the other thing I was thinking, is it would be nice to have docker-gen do a simple http check against the container and not notify nginx until it returns a 200.  if it returns a 5xx, then don't add to nginx.  Thoughts?

fazy:
It seems to me that docker-gen already does two things (simplifying a bit):

1. Watch Docker containers for events (start, stop)
2. Generate config file (and call any kind of reload method as needed)

The issue is when step 2 happens too soon after step 1. We need an intermediate step. This could be, for example:

1. Wait a certain length of time
2. Check a socket (TCP/UDP port *n*)
3. Check a service (HTTP 200 OK)
4. Custom check

You could argue that docker-gen should not take on this responsibility, and if so then the last one, custom check, might be the best option.

I'd be happy for this feature to be implemented like this:

Custom check = a specified command executable. The command is expected to wait for whatever condition is needed to satisfy that the newly created Docker container is fully running. After some time, it should return exit code 0, and then docker-gen can get on with generating the config file.

A non-zero exit code would cause docker-gen to not load the config.

I guess docker-gen would need to call the custom command with some details about the docker container, IP address for sure, maybe port.

Is this likely to work?

jwilder:
I think #67 is general enough that you could implement a delay, HTTP poll, socket test, config check, etc.

fazy:
Yes, that would probably work in most cases.

However, I'm not sure what happens if you write the config file first, as suggested in #67. If two containers start, and one takes a second to become "ready" and the other takes 5 seconds, then the config update/reload would, for a short time, activate the config for the slower container before it's ready.

jwilder:
Fixed via #156

PR #156, opened 2016-01-25
Add -wait parameter to debounce watched events

gigablah:
Addresses the following issues: #42, #43, #57, #103, #136

This is a rewrite of `generateFromEvents` that makes the following changes:

* Only configs that have `watch = true` will be generated upon container events.
* All configs will be regenerated when the events channel is reattached, rather than upon docker client re-initialization. (avoids missing out on events in between)
* Each config with `watch` now has its own separate channel. This is to enable individual debouncing parameters, specified with the `wait` parameter. Syntax is `min:max`, e.g. `500ms:2s`. This feature is referenced from `hashicorp/envconsul`

I've also added an integration test for `generator.go` that verifies the `wait` parameter :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,6 +113,8 @@ Options:
       show version
   -watch
       watch for container changes
+  -wait
+      minimum (and/or maximum) duration to wait after each container change before triggering
 
 Arguments:
   template - path to a template to generate
@@ -153,6 +155,9 @@ path to a template to generate
 watch = true
 watch for container changes
 
+wait = "500ms:2s"
+debounce changes with a min:max duration. Only applicable if watch = true
+
 
 [config.NotifyContainers]
 Starts a notify container section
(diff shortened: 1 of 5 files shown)


gigablah:
A few more changes:

* I made `generateAtInterval` and `generateFromEvents` aware of termination signals as well. (Replaced an apparently unused `quit` channel in `generateAtInterval` in the process)
* Made the docker daemon reconnection logic more robust, so now it will survive a docker restart
* Also added an internal `retry` flag for the generator which determines if it will attempt to reconnect if the events channel is closed, mainly for testing purposes.

jgallen23:
this is awesome! This solves all my issues and would let me safely run docker-gen in prod and give me nearly zero-downtime deployments

jwilder:
Nice @gigablah! Thanks!

PR #158, opened 2016-02-02
Add CurrentContainerID into Docker struct

baptistedonaux:
Add CurrentContainerID into Docker struct. This improvement allow the possibility to compare current container networks and others containers networks.

@md5 I watched your comments and I edited my code to propose a best version.

Linked with https://github.com/jwilder/nginx-proxy/pull/337

### DIFF
--- a/context.go
+++ b/context.go
@@ -32,15 +32,15 @@ func SetServerInfo(d *docker.Env) {
 	mu.Lock()
 	defer mu.Unlock()
 	dockerInfo = Docker{
-		Name:                 d.Get("Name"),
-		NumContainers:        d.GetInt("Containers"),
-		NumImages:            d.GetInt("Images"),
-		Version:              dockerEnv.Get("Version"),
-		ApiVersion:           dockerEnv.Get("ApiVersion"),
-		GoVersion:            dockerEnv.Get("GoVersion"),
-		OperatingSystem:      dockerEnv.Get("Os"),
-		Architecture:         dockerEnv.Get("Arch"),
-		CurrentContainerID:   GetCurrentContainerID(),
+		Name:               d.Get("Name"),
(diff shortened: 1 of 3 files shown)


baptistedonaux:
Success Travis !

md5:
This is looking pretty good. I'm curious about https://github.com/baptistedonaux/docker-gen/commit/b3b09d7553473113c4d935aed720dfb42f560359 though. If different Docker versions have a different name for the cgroups, it might be good to support both formats. I've only seen the slash version looking at Docker 1.9 and 1.10.

I still think this code needs to have a test or two.

Also, I'd recommend initializing the regex as a `const` instead of compiling it each time the function runs.

baptistedonaux:
About https://github.com/baptistedonaux/docker-gen/commit/b3b09d7553473113c4d935aed720dfb42f560359 I had Docker 1.9.1 and works on Debian Jessie. Unfortunately I'm not reproduce.

I change my code with your comments. For the test, and test in code or unit test ?

baptistedonaux:
Travis success. Ready to be merged ! @jwilder @md5

context.go
```diff
@@ -139,12 +143,46 @@ type Mount struct {
...
+	Architecture       string
+	CurrentContainerID string
+}
+
+func GetCurrentContainerID() string {
+	file, err := os.Open("/proc/self/cgroup")
+
+	if err != nil {
+		if os.IsExist(err) {
+			log.Printf("Fail to open /proc/self/cgroup: %s\n", err)
```
md5: This log line is going to make `docker-gen` noisy when it runs outside a container. I don't think it's necessary.

baptistedonaux: Edited !

jwilder:
Thanks @baptistedonaux @md5!

PR #164, opened 2016-02-19
allow dotted map keys in deepGet

codekitchen:
The primary use case here is to allow calling groupBy with a label,
by convention labels are prefixed with a reverse domain name such as
"com.docker.compose".

I'm not in love with the syntax, I'm open to suggestions on what to do there. Technically any syntax is not 100% backwards compatible, since a map key could be any arbitrary string. I'm hoping a backslash at the end of a map key is unusual enough that this is OK.

### DIFF
--- a/README.md
+++ b/README.md
@@ -338,6 +338,8 @@ For example, this is a JSON version of an emitted RuntimeContainer struct:
 
 #### Functions
 
+`$fieldPath` arguments: A field path expression is a dot-delimited list of map keys or struct member names specifying the path from container to a nested value, which must be a string. Dots in map keys can be escaped with a backslash, which must also be escaped, such as `"my\\.dotted\\.key"`.
+
 * *`closest $array $value`*: Returns the longest matching substring in `$array` that matches `$value`
 * *`coalesce ...`*: Returns the first non-nil argument.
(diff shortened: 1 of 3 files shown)


md5:
Did you try using the built-in `index` function? e.g. `index .Labels "com.docker.compose.service"`?

codekitchen:
Yeah, this is specifically for the $fieldPath parameter to the `groupBy` and `where` family of functions, where I can only pass in a dot-delimited string not use a function such as `index`

md5:
For the `where` use cases, you can likely use the `whereLabel*` functions introduced in #143. Adding a `groupByLabel` would be pretty straightforward.

codekitchen:
Sure, I can add a `groupByLabel` function instead of you prefer, and that'll solve my use case. I don't have strong feelings either way, I just figured fewer functions was better.

md5:
It's up to @jwilder, but I would personally prefer a `groupByLabel`.

codekitchen:
Yeah no worries, I was in the code anyway so I pushed up that option as well https://github.com/jwilder/docker-gen/pull/165

jwilder:
Closing this in favor of #165

PR #165, opened 2016-02-19
add groupByLabel function

codekitchen:
This is convenient, but also necessary since the `$fieldPath` parameter to `groupBy` can't contain dots in map key names, and labels are typically reverse-domain-name prefixed.

### DIFF
--- a/template.go
+++ b/template.go
@@ -116,10 +116,10 @@ func groupByKeys(entries interface{}, key string) ([]string, error) {
 func groupByLabel(entries interface{}, label string) (map[string][]interface{}, error) {
 	getLabel := func(v interface{}) (interface{}, error) {
 		if container, ok := v.(RuntimeContainer); ok {
-			if container.Labels[label] == "" {
-				return nil, nil
+			if value, ok := container.Labels[label]; ok {
+				return value, nil
 			}
-			return container.Labels[label], nil
+			return nil, nil
 		}
 		return nil, fmt.Errorf("Must pass an array or slice of RuntimeContainer to 'groupByLabel'; received %v", v)
 	}
--- a/template.go
+++ b/template.go
@@ -51,7 +51,7 @@ func getArrayValues(funcName string, entries interface{}) (*reflect.Value, error
 }
 
 // Generalized groupBy function
-func generalizedGroupBy(funcName string, entries interface{}, key string, addEntry func(map[string][]interface{}, interface{}, interface{})) (map[string][]interface{}, error) {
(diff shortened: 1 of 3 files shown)


template.go
```diff
@@ -102,6 +112,22 @@ func groupByKeys(entries interface{}, key string) ([]string, error) {
 	return ret, nil
 }
 
+// groupByLabel is the same as groupBy but over a given label
+func groupByLabel(entries interface{}, label string) (map[string][]interface{}, error) {
+	getLabel := func(v interface{}) (interface{}, error) {
+		if container, ok := v.(RuntimeContainer); ok {
+			if container.Labels[label] == "" {
```
md5: This will break if the label actually has a blank value. You may want to check the second return value from `container.Labels[label]` to see if it is `false` instead of checking for `""`. I did this in the `whereLabel` function.

codekitchen: Ah you are right, I misread `groupBy` as excluding the container if the value is blank, so I went this route for consistency, but it only excludes if the value is nil. I'll change it.

codekitchen: Done

tompson:
I built something specific for grouping by the service name of the docker compose file:

https://github.com/troii/docker-gen/commit/e9eded7a42463db6014b33ca1bdfcb6c305c97e4

but this pull would make mine obsolete and make it more generic

my vote for this one! ;)

tompson:
I tried to use groubByLabel like 

    {{ range $host, $containers := groupByLabel $ "com.docker.compose.service"  }} 

but I always get

    function "groupByLabel" not defined

codekitchen:
@tompson I've successfully used the new function with a binary built by running `make` with this PR applied, e.g. `make && ./docker-gen -only-exposed /path/to/my/tempate.tmpl`. It sounds like you're probably not using the right docker-gen binary?

tompson:
did exactly that, checked multiple times that I got the right binary

do you have an example template file?

codekitchen:
You probably don't have this PR applied locally then. A simple example template would be:

```
{{range $project, $containers := groupByLabel $ "com.docker.compose.project"}}{{printf "%s %s\n" $project (first $containers).ID}}{{end}}
```

tompson:
wow, that was hard to track down, problem was that in `cmd/docker-gen/main.go` the import 

    "github.com/jwilder/docker-gen"

caused that make did not build `codekitchen/docker-gen` or `troii/docker-gen` even when I ran make in my repository

codekitchen:
Ah dang, that's a pain yeah. Glad you got it sorted out.

md5:
@tompson When I work on this project I add an additional Git remote to my checkout of `github.com/jwilder/docker-gen` to avoid this problem.

jwilder:
Thanks @codekitchen!

PR #175, opened 2016-03-24
Documentation: fix signal number/name mismatch

sherter:
SIGHUP = 1
SIGINT = 2

See output of 'kill -l'


jwilder:
Thanks @sherter!

PR #192, opened 2016-05-19
Add environment variables templates

ldnunes:
With this change it becomes possible for containers to setup it's own configuration template. By default it is disabled.

### DIFF
--- a/cmd/docker-gen/main.go
+++ b/cmd/docker-gen/main.go
@@ -30,6 +30,8 @@ var (
 	configs                 dockergen.ConfigFile
 	interval                int
 	keepBlankLines          bool
+	useEnvVarTemplate       bool
+	destination             string
 	endpoint                string
 	tlsCert                 string
 	tlsKey                  string
@@ -50,7 +52,7 @@ func (strings *stringslice) Set(value string) error {
 }
 
 func usage() {
-	println(`Usage: docker-gen [options] template [dest]
+	println(`Usage: docker-gen [options] [template] [dest]
 
 Generate files from docker container meta-data
 
@@ -59,8 +61,10 @@ Options:`)
 
 	println(`
 Arguments:
-  template - path to a template to generate
-  dest - path to a write the template.  If not specfied, STDOUT is used`)
+  template - path to a template to generate.  If not specfied, it is possible to
+             set the template as a string directly using the
+             VIRTUAL_HOST_TEMPLATE variable on the containers.
+  dest - path to a write the template.  Also as "-destination" option.  If not specfied, STDOUT is used.`)
 
 	println(`
 Environment Variables:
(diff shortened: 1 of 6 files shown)


SergiuCip:
Thanks a bunch! :+1: I really appreciate your effort for doing this. I will give it a try and let you know, however, by looking at the code it is what I was after :)

arnested:
This looks very interesting for a setup I have in mind.

Is it correct that when both the template argument _and_ the environment variable is set the priority is to use the environment variable (at least that's how I read the code -- and what would make sense to me)?

ldnunes:
@arnested Exactly. With this change if the container is started with the template environment variable set then it is used. If not the template set as parameter of the docker-gen is used. We've been used this for some months at my company and looks stable.

Also, we've been planning on adding another feature to "group" the containers with the same template so that docker-gen can put them on the same grouped template, and do things like NGINX upstream server with multiple addresses. I've just need to find some time to implement this... :)

ldnunes:
@arnested Just pushed the "grouping" feature I talked about. Please have a look, if you're still interested. :)

copernico:
Any news?

ldnunes:
I was thinking about this change and I noticed that the variable name VIRTUAL_HOST_TEMPLATE seems too coupled with the nginx-proxy project. I'll edit the changes and instead use something like DOCKERGEN_CUSTOM_TEMPLATE, as soon as I find time to do so.

arnested:
Hi @ldnunes,

I finally got around to giving your code a test ride, but I'm stuck on building it:
```
$ make
Building docker-gen
# _/home/arne/go/src/github.com/jwilder/docker-gen/cmd/docker-gen
cmd/docker-gen/main.go:155: unknown field 'UseEnvVarTemplate' in struct literal of type dockergen.Config
cmd/docker-gen/main.go:173: config.UseEnvVarTemplate undefined (type dockergen.Config has no field or method UseEnvVarTemplate)
Makefile:10: recipe for target 'docker-gen' failed
make: *** [docker-gen] Error 2
```

arnested:
Never mind ... I forgot `go get ./...` after I switched to your branch ....

arnested:
First impression.

Running `docker-gen` without options will normally result in the output of `--help`.

Your branch will output an error message:
```
2017/05/06 13:07:32 Error: missing template file argument or the "use environment variable template" option parameter or configuration has not been set.
```
Probably some kind of check implemented too early in the process?

arnested:
Second impression.

It seems to make no difference whether I include the `-use-environment-variable-template` option or not (or if I set it to `-use-environment-variable-template=false`.

The `VIRTUAL_HOST_TEMPLATE ` variable always affects the generated file.

arnested:
Third impression.

I need to provide the template _in_ the environment variable? Not let the environment variable refer to a template file?

I can see how this is necessary given we don't know if docker-gen is running inside a docker container itself and cannot be guaranteed to see the template file.

On the other hand shuffling a complete template into a environment variable is not easy and does not increase readability of your docker-compose file or whatever you use to start your containers.

jwilder:
Thanks for the PR.  I'm going to close this as this is not a feature I'd like to support currently.

moomoo-dev:
Is there any reason why we can't get this pushed? I would like to get container ENV from the new container that I using to build the container with.

PR #202, opened 2016-06-17
Handle kill event

dustinblackman:
Docker's `kill` event is emitted when `docker stop ID` is executed. In a real world scenario where this becomes very useful is when loading balancing between containers using [nginx-proxy](https://github.com/jwilder/nginx-proxy). If the application within the container is watching for `SIGTERM`, it can instead hold itself open for any specific amount of time to finish in progress connections. NGINX on the other hand will remove the container from it's configuration to prevent any new requests from coming in, but will still hold the open connections until they finish (or when the container gives up and shuts down). 

Docker at the moment doesn't have a `State` variable to define when a container is in progress of shutting down. In this case I've added `Killing` to the `State` struct that's defaulted to false that only flips to true when the `kill` event is emitted. Then is handle when generating the file.


dustinblackman:
@jwilder Any comments on this one? :)

PR #208, opened 2016-08-06
Adds TLS reverse proxies example

pierreozoux:


buchdag:
That was 6 years ago, not sure it's relevant now that we have acme-companion.

PR #212, opened 2016-09-19
add basic arithmetic template functions

byxorna:
I found these functions useful to have for doing (admittedly brittle) logic in my views, when splitting strings and extracting ranges from arrays.


byxorna:
@jwilder not sure if HEAD's build is broken, or the travis failures are expected.

PR #237, opened 2017-02-14
Add build for armhf on alpine

benediktarnold:
Added an armhf build for alpine to the list of builds. Would be nice to download the binary directly from GitHub releases.


jwilder:
@benediktarnold I'll add these for the next releases.

PR #239, opened 2017-02-17
Swarm mode support

s4s0l:
1) parameter -swarm makes docker-gen look for containers in tasks not containers
2) go-dockerclient version changed to 50724541fde5c77c754caf1d65d3c57e3e91883b

How it works: when -swarm switch is on, and -endpoint is a swarm manager node RuntimeContainer struct is created from tasks not containers. Some metadata are not available in swarm tasks so some information is missing but majority of important ones are present (ips, labels, envs, networks) so it is somewhat compatible with templates created for regular mode.

One dirty trick done is that there is no option in swarm mode to expose port without exposing it to ingres network, so if task has no published ports an attempt is made to check exposed port in container image, which may not always work, as image may not be present on endpoint docker-gen is connected.

What do you think?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: go
 go:
-- 1.4
+- 1.7.5
 install:
 - make get-deps
 script:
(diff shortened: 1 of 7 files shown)


s4s0l:
I'm no go expert but this build seems to be failing because travis has go in version 1.4, and go-dockerclient needs something newer...

s4s0l:
Sorry for the mess in commits. If it's a problem I can file PR again with one squashed.

s4s0l:
more i test it, more i hate it

PR #241, opened 2017-03-11
Add HostnameFull Variable to Template Context

zanbaldwin:
Add a variable called `HostnameFull` to the template context since the `Hostname` variable supplied by docker is missing the TLD (and the TLD supplied in Docker's `Domainname` variable isn't included in the context).

An alternative could be to add the `Domainname` variable instead, which would be closer to the information provided by Docker. The decision to provide `HostnameFull` is personal preference only.

Means you can do things like specify the hostname instead of having to specify `VIRTUAL_HOST` environment variables, or other crazy things like automatically create SSH login configs:

```
{{ range $i, $container := . }}
{{ range $j, $address := $container.Addresses }}
{{ if eq $address.Port "22" }}
Host {{ $container.Hostname }}
    HostName {{ $container.HostnameFull }}
    User docker
    IdentityFile ~/.ssh/id_rsa
    Port {{ $address.HostPort }}
{{ end }}
{{ end }}
{{ end }}
```

### DIFF
--- a/generator.go
+++ b/generator.go
@@ -378,6 +378,7 @@ func (g *generator) getContainers() ([]*RuntimeContainer, error) {
 			},
 			Name:         strings.TrimLeft(container.Name, "/"),
 			Hostname:     container.Config.Hostname,
+			HostnameFull: container.Config.Hostname + "." + container.Config.Domainname,
 			Gateway:      container.NetworkSettings.Gateway,
 			Addresses:    []Address{},
 			Networks:     []Network{},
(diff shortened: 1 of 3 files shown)


generator.go
```diff
@@ -378,6 +378,7 @@ func (g *generator) getContainers() ([]*RuntimeContainer, error) {
 			},
 			Name:         strings.TrimLeft(container.Name, "/"),
 			Hostname:     container.Config.Hostname,
+			HostnameFull: container.Config.Hostname + "." + container.Config.Domainname,
```
jwilder: Should this check that `Domanname` is not `""` before joining w/ `"."`?  If it's `""`, `HostnameFull` could get a value like `foo.`

jwilder: Yes, adding `Domainname` would be preferred.

PR #243, opened 2017-04-24
Creates an example dockergen plist

genevera:
An example launchd plist file that I use along with [corectl](http://github.com/TheNewNormal/corectl) to let docker-gen monitor docker and restart dnsmasq.  This is essentially an de-containerized hack of [docker-dns-gen](https://github.com/jderusse/docker-dns-gen) and uses the base templates from that utility.

Not sure if this is too involved to make a good example or not.  I figured someone else out there might appreciate it either way.


josegonzalez:
@buchdag should the template examples just get merged, or is it maybe better to just close them vs have a random assortment of examples here?

PR #247, opened 2017-07-10
Add support for notifying a service

njdancer:


k3a:
It works great, thanks. I don't understand why this hasn't been merged already.

I would just add this at the beginning of task iteration body:
```go
			if task.Status.State != "running" {
				continue
			}
```

Issue #177, opened 2016-03-28
Go Template seems to be a bit limited. Is there any way we might have e.g. jinja2?


ghost:
Go Template'ing seems to be limited and makes some rather basic logical flows impossible or very hard to pull off, see here why: https://github.com/golang/go/issues/14992

While in other templating languages the same logical flow isn't trivially doable either, they usually provide possible workarounds that allow a logical flow as in the link above (for example in jinja2: http://stackoverflow.com/questions/4870346/can-a-jinja-variables-scope-extend-beyond-in-an-inner-block ).

Have you therefore considered providing a more powerful templating language like the popular jinja2 as an alternative? I am aware this would introduce an additional python dependency, but right now I can't do many things in the template I would like to do because of the limitations of Go Template...

JonasT:
As has been pointed out, the specific use case I gave is doable thanks to various additions of docker-gen for the available functions.

Still, I wonder if some templating engine that is both more popular and has a richer set of known default functions for people might at least be worth a consideration? If it's feasible at all to add it as an alternative of course..

PR #254, opened 2017-10-01
Add ability to use pongo2 as an alternative to Go Template

fdintino:
This is more a proposal than a finished pull request—I wanted to gauge interest and get feedback on the implementation before I started writing tests and updating documentation.

[pongo2](https://github.com/flosch/pongo2) is a Go template engine that has a Django/Jinja2-style syntax. My interest in having this available as an option—besides personal preference—is that my development team at The Atlantic is very familiar with the syntax of jinja2, but less so with Go templates. It seems like there might also be others interested in this as well (e.g. #177).

In addition to adding the ability to use pongo2 as the template engine (either with the `-engine pongo2` CLI flag or with `engine = "pongo2"` in a config file) I've ported the existing example templates to pongo2.

I was tempted to make the functions and context variables all lowercase / underscore-separated to make them more idiomatic for Django-style templates, but I thought that might be a bridge too far.


fdintino:
@jwilder any feedback?

jwilder:
@fdintino I like it.  I need to find some time to test it out, but looks pretty good at first glance.

PR #264, opened 2017-12-01
feat: Add "whereNot" function

### LINKED ISSUE #261: where function for non-matches/false
Is there an inverse for the `where` function to return non-matches, something like `whereNot`?
I'm trying to filter out certain containers rather than include specific containers based on their `Image.Registry`

Thiht:
Fix #261 

The goal is to filter out items not matching a property. My use case was to filter out non-published Addresses (ie. with `HostPort == ""`).

This PR allows me to solve it with:

```go
{{ $addresses := whereNot $container.Addresses "HostPort" "" }}
```


jwilder:
Thanks @Thiht!

Thiht:
Thank you :)

Is it possible for you to make a [new release](https://github.com/jwilder/docker-gen/releases) soon in order for the Docker image to be updated?

PR #303, opened 2019-02-11
Add Labels to SwarmNode

xdefrag:


aido93:
In the vanilla image it is not supported to create DNS names of containers based on the swarm node labels. Lets imagine situation: we have container that exposes 2 ports: one of them is just working port (like 80), and the other is exporter port (for instance, 9811). Also we need to use swarm proxying only for 80 port but exporters must have unique dns names per machine without swarm load balancing. We just need to know where response come from. So this pull request resolves this use case.

xdefrag:
@jwilder can you review this PR?

PR #308, opened 2019-03-18
Make sure file generated by docker-gen has the right permissions

NicolasDorier:
If the destination file does not exists, the permissions become `600` because this is the default permission for temp files.

This is problematic, if the container consuming the destination file does not run as root.

Instead, the default permissions should be used (as per umask).
This PR will make permission `644`, which is generally what we want. And if the container consuming it is not happy with it, it always can override the permissions.

### DIFF
--- a/template.go
+++ b/template.go
@@ -517,7 +517,16 @@ func GenerateFile(config Config, containers Context) bool {
 		}
 
 		oldContents := []byte{}
-		if fi, err := os.Stat(config.Dest); err == nil {
+		if fi, err := os.Stat(config.Dest); err == nil || os.IsNotExist(err) {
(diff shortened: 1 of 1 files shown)


gamma:
This issue is still open, right?

buchdag:
I'm going to close then re-open this PR to trigger the tests.

buchdag:
Ack, crap, I forgot this repo was still using Travis CI.

buchdag:
Merged in #352

gamma:
Thanks!!!

buchdag:
If you have other Issues / PR of interest in mind don't hesitate to tag me, I've been mostly focusing my efforts on `nginx-proxy`.

PR #309, opened 2019-03-18
Add read function to read the content of a file

NicolasDorier:
This has been very useful in my fork for the following usecase:

I wanted to use docker-gen with TOR. When you create a hidden service, TOR create a file at a specific location with the TOR onion host that represent your service. I needed to use this onion host as input for generating my nginx configuration file. So my nginx.tmpl needed to read a file.


aywan:
Two years later i did almost same thing https://github.com/nginx-proxy/docker-gen/pull/399
The feature not popular, but required in some cases

josegonzalez:
This was implemented [here](https://github.com/nginx-proxy/docker-gen/pull/399) through the `include` function, so I think this one can be closed.

buchdag:
Thanks for for the reminder @josegonzalez 👍

PR #311, opened 2019-05-23
Improve notify-sigup to allow for the use of filters (such as labels)

### LINKED ISSUE #77: Allow -notify-sigup to work with docker-compose
In a docker-compose situation, the container name isn't known in advance.  A way to tell -notify-sigup to pull the container name from an environment variable (ie $nginx_NAME) would be helpful.

After more reading, I might be misunderstanding how docker-compose sets the environment variables.

I'm looking for a solution to putting docker-gen in a separate container from nginx and still being able to use -notify-sigup on the container which gets turned into <projectname>_nginx_1 and not really known ahead of time (unless you assume how docker-compose is going to name the containers).

This might get solved in a future release when docker-compose includes the ability to name your containers whatever you'd like.

tjvdmolen:
This implementation will allow for the use of a filter to live query the containers before sending the sigup, and is fully backwards compatible with the previous implementation.

Resolves #77


buchdag:
Superseded by #619

PR #322, opened 2020-04-02
Use args instead of variables in Dockerfile

### LINKED ISSUE #321: Use args instead of env variables in Dockerfile
By using arg instead of variables in Dockerfile it will be easy to build docker-gen for different architectures, e.g. I can have following docker compose file and it will work correctly:

```
version: '3'
services:
  nginx-gen:
    build:
      context: https://github.com/dealancer/docker-gen.git
      args:
        - ARCHITECTURE=armhf
...
```

dealancer:
This PR fixes #321.


sgabe:
@dealancer Note that `VERSION` is later used in `SetServerInfo()`:
https://github.com/jwilder/docker-gen/blob/4edc190faa34342313589a80e3a736cafb45919b/context.go#L37
You might want to set the environment variable based on the build argument:
```
ARG VERSION=0.7.4
ENV VERSION ${VERSION}
```

PR #324, opened 2020-05-31
Support cgroup docker_limit.slice memory and cpu limits for docker daemon

mickelindahl:
I had to restrict the total resources that the docker daemon can utilize on a server. The solution inlcuded adding a docker_limit.slice file to /etc/systemd/system (see https://github.com/mickelindahl/docker_daemon_limit_total_resources). This results in that in "/proc/self/cgroup" containers are named .../docker_limit.slice/{container id}. Thus matchDockerCurrentContainerID nor matchECSCurrentContainerID could pick up the container name which left me with a broken nginx-proxy. I have added an additional funcition matchDockerLimitCurrentContainerID witch solves this problem. Would love if you could add this to the project.

Cheers
/Mikael

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 docker-gen
 dist
 *.gz
+.idea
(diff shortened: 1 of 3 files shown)


buchdag:
Superseded by #336, #345 and #356

PR #343, opened 2021-04-05
Lift restriction on the function contains

AlexanderLieret:
The function `contains` only allowed to check maps of the type `map[string]string`. This PR lifts this restriction and allows for arbitrary types to be used.

### DIFF
--- a/README.md
+++ b/README.md
@@ -353,7 +353,7 @@ For example, this is a JSON version of an emitted RuntimeContainer struct:
 
 * *`closest $array $value`*: Returns the longest matching substring in `$array` that matches `$value`
 * *`coalesce ...`*: Returns the first non-nil argument.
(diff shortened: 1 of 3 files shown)


buchdag:
@AlexanderLieret maybe the test unit should also be declined for other basic go types like `bool`, `uint` and `float32` ?

AlexanderLieret:
@buchdag I don't have much experience with go. This was a simple test to test what happens if a wrong key datatype is supplied.

I bit of testing revealed that the types of the keys are compared first. Checking with other types should not be necessary.

```
m := map[int]bool { 42: true }
m[42] // true
m[42.0] // true
m["42"] // compiler error
m[m] // compiler error
contains(m, 42) // true
contains(m, 42.0) // false
contains(m, "42") // false
contains(m, m) // false
```

jwilder (approved):

PR #336, opened 2021-01-21
Fix docker-gen breaking on new docker version (See #335)

NicolasDorier:
This PR is scanning for the full container ID by getting the short container id via `HOSTNAME`, then trying to fetch the rest of it in "/proc/self/cgroup" or "/proc/self/mountinfo".

So no breaking change to old versions of docker.

Alternative to #335

Note I did not tested on new docker, but I tested on old one and nothing break.

PR #345, opened 2021-04-09
Fix GetCurrentContainerID() and corresponding test

buchdag:
This PR adds a fall back container ID matching function to the hostname based container ID matching function introduced in #336, in order to fix the regression with containers using a custom hostname.

The `GetCurrentContainerID()` function will try the hostname based container ID matching function on all files before using the fallback function in case it didn't get a valid container ID. Its signature has been altered to allow testing it with real files reads instead of testing only the matching functions. This slightly increased test coverage from 62.1 % to 64.2 %.

`/proc/1/cpuset` has been added as a third alternative to `/proc/self/cgroup` and `/proc/self/mountinfo` (it's one of the container ID source we use on nginx-proxy/acme-companion).

`matchDockerCurrentContainerID()` has been renamed to `matchContainerIDWithHostname()` for clarity. Claiming it matches the current container ID feels a bit untrue, as it just match a 64 characters hex starting with 12 characters HOSTNAME.

### DIFF
--- a/context.go
+++ b/context.go
@@ -2,11 +2,14 @@ package dockergen
 
 import (
 	"bufio"
+	"bytes"
+	"fmt"
+	"io"
 	"os"
 	"regexp"
 	"sync"
-	"fmt"
-	"github.com/fsouza/go-dockerclient"
+
+	docker "github.com/fsouza/go-dockerclient"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


buchdag:
Forgot that: the `matchECSCurrentContainerID()` function has been removed entirely has ECS changed its cgroup format a long time ago and it wasn't matching anything anymore.

jwilder (approved):

jwilder:
Nope.   I hit the wrong approve option.

PR #346, opened 2021-04-09
Have the Dockerfile build from the actual surrounding sources rather than downloading a hardcoded version

buchdag:
While I understand the rationale for downloading a hard coded version back when building the Go binary using multi stage Dockerfile wasn't an option, I think it's not relevant anymore and that the Dockerfile should build an image corresponding to the surrounding sources. We're currently in a strange situation where the project's Dockerfile potentially builds an image that use codes dating from a handful of commits back.

Having tagged versions of the container derived from git tags should happen through the CI  instead (either GitHub Actions of DockerHub's autobuild). The version can still be passed to `go build -X main.buildVersion` using the `VERSION` build `ARG`.

The resultant image has been tested with nginx-proxy and is working.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,28 +1,27 @@
 # Build docker-gen from scratch
-FROM golang:1.14-alpine as dockergen
-RUN apk add --no-cache git
+FROM golang:1.16-alpine as go-builder
 
-# Download the sources for the given version
-ENV VERSION 0.7.5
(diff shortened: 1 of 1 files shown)


jwilder (approved):

buchdag:
@jwilder unfortunately DockerHub can't do automated build for multiple archs. Do you want me to contribute a GitHub Actions worklow that will take care of that by building on GHA with buildx and pushing to DockerHub ? It'll only require that you create a token on DockerHub and add it to the GitHub repository's secrets.

PR #348, opened 2021-04-13
Explicitly make docker socket rw in documentation

Syphdias:
Docker does not allow for read-only mounts of sockets at this point.
This can be tested by running:
```
❯ docker run --rm -it -v /var/run/docker.sock:/tmp/docker.sock:ro \
    docker ls -l /tmp/docker.sock
srw-rw----    1 root     967              0 Apr 13 16:16 /tmp/docker.sock
❯ docker run --rm -it -v /var/run/docker.sock:/tmp/docker.sock:ro \
    docker /usr/local/bin/docker -H unix:///tmp/docker.sock ps |wc -l
17
```

Even if docker supported read-only sockets, you would not want it to be
read-only if you are using `-notify-sighup` because then the docker socket is
used to communicate to the other container to notify it.

This change in documentation reflects the fact that the socket will be writable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ Fetch the template and start the docker-gen container with the shared volume:
 $ mkdir -p /tmp/templates && cd /tmp/templates
 $ curl -o nginx.tmpl https://raw.githubusercontent.com/jwilder/docker-gen/master/templates/nginx.tmpl
 $ docker run -d --name nginx-gen --volumes-from nginx \
-   -v /var/run/docker.sock:/tmp/docker.sock:ro \
+   -v /var/run/docker.sock:/tmp/docker.sock:rw \
    -v /tmp/templates:/etc/docker-gen/templates \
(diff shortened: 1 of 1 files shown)


buchdag:
While `:ro` does not prevent writes to the socket, I think it does prevent the socket's deletion by the container.

Syphdias:
I don't think that's possible with `:rw` either:
```
❯ docker run --rm -it -v /var/run/docker.sock:/tmp/docker.sock:rw docker rm /tmp/docker.sock
error during connect: Delete http://docker:2375/v1.24/containers/tmp/docker.sock: dial tcp: lookup docker on 10.2.23.83:53: server misbehaving
❯ docker run --rm -it -v /var/run/docker.sock:/tmp/docker.sock:rw docker sh
/ # rm /tmp/docker.sock
rm: can't remove '/tmp/docker.sock': Resource busy
/ # 
```

PR #350, opened 2021-04-30
Fix problem with docker `internal` networks

anlorn:
Hello, I found an issue while setting up `nginx-proxy` container for my network.

I have the following setup:
1) network 'A' is a regular network created in docker
2) network 'B' is an internal network created with `--internal` parameter

Container with `nginx-proxy` connected to both `A` and `B`  networks. At the same time, containers with applications are connected only to 'B' network. When running such a config, I was always getting.

After investigating, I find out that the issue comes from the fact that 'docker inspect' section `NetworkSettings.Ports` is always empty for internal networks, no matter which port you expose/publish. As the result we always have empty `addresses` variable when we generate a template. Solution here is to take exposed ports from section `Config.ExposedPorts` when `NetworkSettings.Ports` is empty.

How to reproduce:
Create internal network `docker network create 'test_network' --internal`
Run httpd container connected on this network ` docker run --network test_network --rm -e VIRTUAL_HOST=test.local -e VIRTUAL_PORT=80 httpd:2.4-alpine`
Get latest nginx.tml `wget https://raw.githubusercontent.com/nginx-proxy/nginx-proxy/main/nginx.tmpl`
Run `docker run --network test_network -it --rm -v /run/docker.sock:/tmp/docker.sock -v $(pwd):/app/ jwilder/docker-gen /app/nginx.tmpl`
You will see that in the `upstream` section of the generated config `server` directive is marked as `down`.
Repeat all the steps but this time, on step 1, create a network without the `--internal` parameter.

### DIFF
--- a/generator.go
+++ b/generator.go
@@ -397,21 +397,34 @@ func (g *generator) getContainers() ([]*RuntimeContainer, error) {
 			IP6LinkLocal: container.NetworkSettings.LinkLocalIPv6Address,
 			IP6Global:    container.NetworkSettings.GlobalIPv6Address,
 		}
-		for k, v := range container.NetworkSettings.Ports {
-			address := Address{
-				IP:           container.NetworkSettings.IPAddress,
-				IP6LinkLocal: container.NetworkSettings.LinkLocalIPv6Address,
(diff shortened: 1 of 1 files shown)


buchdag:
Superseded by #579

PR #352, opened 2021-06-15
Make sure file generated by docker-gen has the right permission

buchdag:
#308 by @NicolasDorier, with a minor code formatting correction

> If the destination file does not exists, the permissions become `600` because this is the default permission for temp files.
> 
> This is problematic, if the container consuming the destination file does not run as root.
> 
> Instead, the default permissions should be used (as per umask).
> This PR will make permission `644`, which is generally what we want. And if the container consuming it is not happy with it, it always can override the permissions.

### DIFF
--- a/template.go
+++ b/template.go
@@ -517,7 +517,16 @@ func GenerateFile(config Config, containers Context) bool {
 		}
 
 		oldContents := []byte{}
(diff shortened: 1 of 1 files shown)


NicolasDorier:
Damn 2 years ago. I don't remember I even opened this PR, nor that I did write anything in the previous PR. :D

PR #353, opened 2021-06-16
Sorting, reversing and reverse sorting

gamma:
The PR adds sorting, reversing and reverse sorting of object lists. I was using these function in my fork for some time, since docker-gen looked stale some time ago. Would be nice to add these functions.

@buchdag can you please have a look at it?

### DIFF
--- a/template.go
+++ b/template.go
@@ -14,6 +14,7 @@ import (
 	"path/filepath"
 	"reflect"
 	"regexp"
+	"sort"
 	"strconv"
 	"strings"
 	"syscall"
@@ -428,6 +429,36 @@ func when(condition bool, trueValue, falseValue interface{}) interface{} {
 	}
 }
 
+// sortStrings returns a sorted array of strings
+func sortStrings(values []string) []string {
+	sort.Strings(values)
+	return values
+}
+
+// sortStrings returns a sorted array of strings
+func sortStringsReverse(values []string) []string {
+	sort.Sort(sort.Reverse(sort.StringSlice( values )))
+	return values
+}
+
+// sortObjects returns a sorted array of objects (sorted by object key field)
+func sortObjects(objs interface{}, key string) (interface{}, error) {
+	objsVal, err := getArrayValues("sortObj", objs)
+	if err != nil {
+		return nil, err
+	}
+	data := make([]interface{}, objsVal.Len())
+	for i := 0; i < objsVal.Len(); i++ {
+		data[i] = objsVal.Index(i).Interface()
+	}
+	sort.Slice(data, func(i, j int) bool {
+		a := reflect.ValueOf(deepGet(data[i], key)).Interface().(string)
+		b := reflect.ValueOf(deepGet(data[j], key)).Interface().(string)
+		return a < b
+	})
+	return data, nil
+}
+
 func newTemplate(name string) *template.Template {
 	tmpl := template.New(name).Funcs(template.FuncMap{
 		"closest":                arrayClosest,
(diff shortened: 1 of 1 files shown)


buchdag:
@gamma sorry for the delay, that looks nice and might be helpful, would you mind adding tests ?

gamma:
Sure should 😄
I'm not sure though when I can do it. Please keep the PR open until then.

buchdag:
👍 

Do you mind if I add test myself if I find the time at some point ?

gamma:
Absolutely not. Please go ahead 🙏

buchdag:
@gamma I'll probably go in a slightly different direction for `sortObjectsByKeys()` in order to have a bit more extensibility:

```golang
type sortable interface {
	sort.Interface
	set(string, interface{}) error
	get() []interface{}
}

type sortableData struct {
	data []interface{}
}

func (s sortableData) get() []interface{} {
	return s.data
}

// method required to implement sort.Interface
func (s sortableData) Len() int { return len(s.data) }

// method required to implement sort.Interface
func (s sortableData) Swap(i, j int) { s.data[i], s.data[j] = s.data[j], s.data[i] }

type sortableByKey struct {
	sortableData
	key string
}

func (s *sortableByKey) set(funcName string, entries interface{}) (err error) {
	entriesVal, err := getArrayValues(funcName, entries)
	if err != nil {
		return
	}
	s.data = make([]interface{}, entriesVal.Len())
	for i := 0; i < entriesVal.Len(); i++ {
		s.data[i] = reflect.Indirect(entriesVal.Index(i)).Interface()
	}
	return
}

// method required to implement sort.Interface
func (s sortableByKey) Less(i, j int) bool {
	values := map[int]string{i: "", j: ""}
	for k := range values {
		if v := reflect.ValueOf(deepGet(s.data[k], s.key)); v.Kind() != reflect.Invalid {
			values[k] = v.Interface().(string)
		}
	}
	return values[i] < values[j]
}

// Generalized SortBy function
func generalizedSortBy(funcName string, entries interface{}, s sortable, reverse bool) (sorted []interface{}, err error) {
	err = s.set(funcName, entries)
	if err != nil {
		return nil, err
	}
	if reverse {
		sort.Stable(sort.Reverse(s))
	} else {
		sort.Stable(s)
	}
	return s.get(), nil
}

// sortObjectsByKeysAsc returns a sorted array of objects, sorted by object's key field in ascending order
func sortObjectsByKeysAsc(objs interface{}, key string) ([]interface{}, error) {
	s := &sortableByKey{key: key}
	return generalizedSortBy("sortObjsByKeys", objs, s, false)
}

// sortObjectsByKeysDesc returns a sorted array of objects, sorted by object's key field in descending order
func sortObjectsByKeysDesc(objs interface{}, key string) ([]interface{}, error) {
	s := &sortableByKey{key: key}
	return generalizedSortBy("sortObjsByKey", objs, s, true)
}
```

The idea is to be be able to
- easily provide both ascending and descending (reverse) sort function
- be able to add other sorting methods with new implementations of the `sortable` interface (working mostly on the get() and Less() methods)

also this check

```golang
if v := reflect.ValueOf(deepGet(s.data[k], s.key)); v.Kind() != reflect.Invalid {
	values[k] = v.Interface().(string)
}
```

Is needed because `reflect.ValueOf(deepGet(s.data[k], s.key)).Interface().(string)` will panic if the target key isn't set.

I think I'll probably open another PR and co-author the commit with you if that's ok with you.

gamma:
Absolutely. Honestly I'm not at all into Go - I just put together what I thought would be working and it was OK for my case. Please feel free to take this direction - it makes sense for future extensions.

PR #356, opened 2021-08-03
Fix container ID extraction from /proc/self/mountinfo file

### LINKED ISSUE #355: $CurrentContainer is null
After a while I updated manually to `:latest` image from docker hub and all my containers ended up with upstream `server 127.0.0.1 down;`.

I copied `nginx.tmpl`, and after a while I found out `{{ $CurrentContainer := where $ "ID" .Docker.CurrentContainerID | first }}` returns `null`, so this [range](https://github.com/nginx-proxy/nginx-proxy/blob/4e3e518936d27d22e5e5e5552d41097d11bb727c/nginx.tmpl#L177) can never set `$server_found` to true.

Change from `$CurrentContainer.Networks` to `$container.Networks` ([from outer range](https://github.com/nginx-proxy/nginx-proxy/blob/4e3e518936d27d22e5e5e5552d41097d11bb727c/nginx.tmpl#L163)) solved the problem.

I am running Debian Bullseye (daily dist-upgraded until Bullseye is stable), and docker-ce vevrsion 5:20.10.7\~3-0\~debian-bullseye from docker.com (i.e. not debian-packaged docker).

Can somebody confirm this behavior? I can make pull request.

buchdag:
This PR fixes #355 

On some systems both `/proc/1/cpuset` and `/proc/self/cgroup` are left empty, leaving only `/proc/self/mountinfo` to retrieve the ID of the container docker-gen is running inside of. This file differs slightly from the two others and the `GetCurrentContainerID()` and `matchContainerID()` functions had to be rewritten to

1.  process each file separately
2. use a different regexp for `/proc/self/mountinfo`

in order to avoir retrieving a 64 characters ID that is not the container ID.

### DIFF
--- a/context.go
+++ b/context.go
@@ -2,9 +2,7 @@ package dockergen
 
 import (
 	"bufio"
-	"bytes"
 	"fmt"
-	"io"
 	"os"
 	"regexp"
 	"sync"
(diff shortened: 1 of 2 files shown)


multicast:
Matching container ID on the `/etc/hostname` in the `mountinfo` file instead of the first (in my case, overlay fs) may help.

buchdag:
That's more or less what the PR does, if `matchContainerID()` identifies the starting pattern of a `/proc/<pid>/mountinfo` files, it preprends `containers/` to the regular expression, so that the function will match one of the lines with the correct ID instead of the first one.

It should work on setups similar to yours but that's unfortunately still a brittle solution without any guaranteed stability.

PR #362, opened 2021-08-05
Re-organize the project structure, change the name and clean the dependencies.

buchdag:
Pretty much what's said in the title, this PR is an attempt at cleaning up the project structure a little bit, and doing away with the indirect dependencies in `go.mod`.


buchdag:
@tkw1536 is there a point to keep `go` version to `1.11` in `go.mod` ?

tkw1536:
Only if we want to allow compilation under old go versions.

buchdag:
To be honest I have absolutely no idea if people are still compiling docker-gen on out of date versions of go.

On the other hand sticking to `1.11` prevents us from using more recent features of the language:

> For packages within the module, the compiler rejects use of language features introduced after the version specified by the go directive.

If this is project is to be actively maintained and given the fact that we release pre built binaries, I think we should rather stick to up to the current date version of go.

Issue #244, opened 2017-05-22
Restart unless stopped not working


orion110217:
I noticed `restart: unless-stopped` is not working in docker-gen. Do I need to use `always` instead? Why is it not working?

m-a-v:
I have the same problem. But also "always" does not work. After restart of the server I have to startup docker-gen manually. How can I find the reason for the exited docker-image? Why does it not restart automatically when I use always?

jebw:
I'm hitting the same issue - docker-gen container + nginx container - both restart=always - if I restart the docker service, nginx comes back up but docker-gen remains stopped with an exit code of 2

Nothing obvious in the logs

Docker start will get the container running again

obi12341:
Yep, exactly the same for me. Possible to fix this issue? It is very annoying in a production environment.

danielkhalilov:
docker update --restart=unless-stopped 7e99ccfb8139    DOES NOT WORK....

gghg1989:
Same issue here, every time the server rebooted or after ran my auto backup program. All other containers would be up except docker-gen.

Giampa:
Same here, is it something with nginx.tmpl?

m-a-v:
Due to problems with letsencrypt-nginx-proxy-companion I removed this component. Since then I didn't have any problems after reboot. Is it possible that letsencrypt-nginx-proxy-companion causes this trouble?

roopemerikukka:
Has anyone found any solutions to this? I'm facing the same problem. If I reboot the server everything gets up and running except `docker-gen` which is exited with `exit status 2`. Related issue: https://github.com/jwilder/docker-gen/issues/201

Michiel-Idgis:
I have the same problem. I assume the docker-letsencrypt-nginx-proxy-companion container most people will use with it, causes the problem. And only when another app/container triggers the letsencrypt container to do it's work. I Have done some tests with different combinations of containers and only when there was a container that triggerd the letsencrypt container, i got a exited (2) after a server-reboot or docker-machine restart. And not always, but around ~75% of the times.
It happens with 0.7.0, 0.7.3 and 0.7.4. Nginx offical image 1.14 and docker-letsencrypt-nginx-proxy-companion:v1.8
Can anybody tell what is outputting the exit code of 2?

copierrj:
We worked around this problem by wrapping docker-gen in a shell script that executes docker-gen in a loop and traps SIGHUP and SIGTERM signals:

```sh
[...]

PR #378, opened 2021-09-21
Fix exit(2) on SIGHUP before fully initialized.

JoelLinn:
docker-gen uses SIGHUP to trigger regeneration.
However, when SIGHUP is fired before docker-gen is fully initialized (and the signal handler was installed), the process will just exit(2).
This fix ignores SIGHUP until the signal handler is installed.

Fixes https://github.com/jwilder/docker-gen/issues/244

### DIFF
--- a/cmd/docker-gen/main.go
+++ b/cmd/docker-gen/main.go
@@ -5,7 +5,9 @@ import (
 	"fmt"
 	"log"
 	"os"
+	"os/signal"
 	"path/filepath"
+	"syscall"
 
 	"github.com/BurntSushi/toml"
 	docker "github.com/fsouza/go-dockerclient"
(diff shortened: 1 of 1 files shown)


JoelLinn:
https://github.com/nginx-proxy/docker-gen/issues/244#issuecomment-622969364

buchdag:
HI @JoelLinn, thanks for catching and fixing this one. Have you been able to test the fix in production ?

JoelLinn:
I have deployed this fix (as my custom build on-top the master tip of that time) over a year ago on about 40 servers and didn’t have any (re-)generation problems since then.

It seems to be the root cause for the issue #244 (although the issue title is pretty generic so there might be all sorts of causes for a phenomenon like it)

I’m happy this project is maintained again ☺️

PR #399, opened 2022-01-21
Nested templates and include function

aywan:
Hi!

I have the task of including file content in the destination. Now I have a draft template in which I first "render" using bash, and then I use result as a template in docker-gen.

First, I added an import function that returns the contents of a file.
Secondly, I added the ability to parse several templates. To do this, instead of one file, specify several through `;`.

I tried to write the documentation, but I think that it must be corrected.

I don’t know how useful this feature will be, but at my case it can be very useful.

#### just example

**header.inc**:
```
# this file is auto generated
```
**author.tmpl**:
```
# created by {{ author.name }}
# email {{ author.email }}
```
**result.tmpl**:
```
{{ include "header.inc" }}
{{ template "author.tmpl" }}
---
some result content
```
**_executing_**: `docker-gen "result.tmpl;author.tmpl"` 
**result**:
```
# this file is auto generated
# created by 
# email 
---
some result content
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -208,6 +208,8 @@ e75a60548dc9 = 1  # a key can be either container name (nginx) or ID
 
 The templates used by docker-gen are written using the Go [text/template](http://golang.org/pkg/text/template/) language. In addition to the [built-in functions](http://golang.org/pkg/text/template/#hdr-Functions) supplied by Go, docker-gen uses [sprig](https://masterminds.github.io/sprig/) and some additional functions to make it simpler (or possible) to generate your desired output.
 
+For parsing several templates, split path with `;` (for example `template = "nginx.tmpl;header.tmpl"`). This makes possible to use go nested templates through standard `template` function.
+
 #### Emit Structure
 
 Within the templates, the object emitted by docker-gen will be a structure consisting of following Go structs:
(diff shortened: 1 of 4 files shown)


josegonzalez:
Can the include here be used to assign the output to a variable?

josegonzalez:
@buchdag it would be great to have this functionality, as it would allow out of band specification of domains to be associated with apps (where an app is a specific label on a collection of docker containers).

josegonzalez:
Actually this could be quite nice if the entire "app" config for a set of containers was stored in a file. You could group all containers by an app label, grab the config from disk, and generate file output as appropriate. If the app config changes, then whatever the thing is that writes the config could SIGHUP docker-gen, allowing the written files to be updated. This would remove the need for rebuilding containers, which is necessary due to Docker not exposing metadata that can be modified.

This wouldn't be as useful for swarm, but one could implement an `http_request` function that returns a response, allowing users to parse the response as "app" config. In any case, that could definitely be done in a separate PR.

README.md
```diff
@@ -208,6 +208,8 @@ e75a60548dc9 = 1  # a key can be either container name (nginx) or ID
 
 The templates used by docker-gen are written using the Go [text/template](http://golang.org/pkg/text/template/) language. In addition to the [built-in functions](http://golang.org/pkg/text/template/#hdr-Functions) supplied by Go, docker-gen uses [sprig](https://masterminds.github.io/sprig/) and some additional functions to make it simpler (or possible) to generate your desired output.
 
+For parsing several templates, split path with `;` (for example `template = "nginx.tmpl;header.tmpl"`). This makes possible to use go nested templates through standard `template` function.
```
buchdag, line 211: We should mention that the "main" template should be first in the list.

josegonzalez:
@buchdag if the only requested change is a readme change, would it be possible to have this merged and then I can update the readme in a follow-up PR?

buchdag:
@josegonzalez sure 👍

buchdag (approved):

josegonzalez:
@buchdag docs here: https://github.com/nginx-proxy/docker-gen/pull/598

PR #418, opened 2022-03-20
Add template functions from sprig

### LINKED ISSUE #177: Go Template seems to be a bit limited. Is there any way we might have e.g. jinja2?

### LINKED ISSUE #293: Add “Set” Function
Sometimes it is required to have a set of something. For example, I want to get a list of unique networks.

```
{{ range $network := set $containers "Networks" }}
   {{ $network.Name }}
{{ end }}
```

In this example, the `$containers` is a list of docker container created by `groupByMulti $ "Labels.lb" ","`. 

The interface could be: `set $containers $fieldPath` (cf. `groupBy`)

Is it possible to provide a function like this or have I overlooked an existing function that could handle this use case?

### LINKED ISSUE #323: Is it possible to get ip?
Today found that **/etc/nginx/conf.d/default** generates by **/app/nginx.tmpl**
Is that possible to get just IP from specific **Container Name**?
like 
```
if (eq containerName "rcloneBackendtest")
 ipAddr := ... do something for getting ip...
end
```

In nginx.tmpl found big function for this but don’t completely understand how it works.
need to get just IP address. 

Can you help?

rhansen:
This PR adds a bunch of useful template functions from the [sprig library](https://masterminds.github.io/sprig/), which makes it easier/possible to introduce more advanced logic in templates.

Fixes #177 by adding useful functions like `append`, `join`, and `set`.

Fixes #293 by enabling templates to use a Go map object as a set (via the `set` and `hasKey` functions).

Fixes #323 by adding a new `getHostByName` function.

The removed functions were removed because sprig offers those same functions with the same behavior. I left functions with the same name but different behavior alone to avoid breaking existing templates.

### DIFF
--- a/README.md
+++ b/README.md
@@ -206,7 +206,7 @@ e75a60548dc9 = 1  # a key can be either container name (nginx) or ID
 
 ### Templating
 
-The templates used by docker-gen are written using the Go [text/template](http://golang.org/pkg/text/template/) language. In addition to the [built-in functions](http://golang.org/pkg/text/template/#hdr-Functions) supplied by Go, docker-gen provides a number of additional functions to make it simpler (or possible) to generate your desired output.
+The templates used by docker-gen are written using the Go [text/template](http://golang.org/pkg/text/template/) language. In addition to the [built-in functions](http://golang.org/pkg/text/template/#hdr-Functions) supplied by Go, docker-gen uses [sprig](https://masterminds.github.io/sprig/) and some additional functions to make it simpler (or possible) to generate your desired output.
 
 #### Emit Structure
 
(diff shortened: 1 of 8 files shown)


buchdag:
Hi. Could you give us a bit more info about what this PR does and how it fixes those three issues ?

rhansen:
Updated the PR description with more context.

buchdag:
I'll look into this ASAP, thanks for the PR !

rhansen:
I added some tests to ensure that the migration to sprig doesn't break any existing templates.

buchdag:
Could you do `go mod tidy` on  1e8b475 or apply those diffs ?

```diff
diff --git a/go.mod b/go.mod
index ca43ca7..3eeed13 100644
--- a/go.mod
+++ b/go.mod
@@ -4,6 +4,7 @@ go 1.17
 
 require (
        github.com/BurntSushi/toml v1.0.0
+       github.com/Masterminds/sprig/v3 v3.2.2
        github.com/fsouza/go-dockerclient v1.7.10
        github.com/stretchr/testify v1.7.1
 )
@@ -12,7 +13,6 @@ require (
        github.com/Azure/go-ansiterm v0.0.0-20210617225240-d185dfc1b5a1 // indirect
        github.com/Masterminds/goutils v1.1.1 // indirect
        github.com/Masterminds/semver/v3 v3.1.1 // indirect
-       github.com/Masterminds/sprig/v3 v3.2.2 // indirect
        github.com/Microsoft/go-winio v0.5.2 // indirect
        github.com/Microsoft/hcsshim v0.9.2 // indirect
        github.com/containerd/cgroups v1.0.3 // indirect
```

```diff
diff --git a/go.sum b/go.sum
index 174d59f..23858fe 100644
--- a/go.sum
+++ b/go.sum
@@ -1361,6 +1361,7 @@ gopkg.in/yaml.v2 v2.2.4/go.mod h1:hI93XBmqTisBFMUTm0b8Fm+jr3Dg1NNxqwp+5A1VGuI=
 gopkg.in/yaml.v2 v2.2.5/go.mod h1:hI93XBmqTisBFMUTm0b8Fm+jr3Dg1NNxqwp+5A1VGuI=
 gopkg.in/yaml.v2 v2.2.8/go.mod h1:hI93XBmqTisBFMUTm0b8Fm+jr3Dg1NNxqwp+5A1VGuI=
 gopkg.in/yaml.v2 v2.3.0/go.mod h1:hI93XBmqTisBFMUTm0b8Fm+jr3Dg1NNxqwp+5A1VGuI=
+gopkg.in/yaml.v2 v2.4.0 h1:D8xgwECY7CYvx+Y2n4sBz93Jn9JRvxdiyyo8CTfuKaY=
 gopkg.in/yaml.v2 v2.4.0/go.mod h1:RDklbk79AGWmwhnvt/jBztapEOGDOx6ZbXqjP6csGnQ=
 gopkg.in/yaml.v3 v3.0.0-20200313102051-9f266ea9e77c/go.mod h1:K4uyk7z7BCEPqu6E+C64Yfv1cQ7kz7rIZviUmN+EgEM=
 gopkg.in/yaml.v3 v3.0.0-20200615113413-eeeca48fe776/go.mod h1:K4uyk7z7BCEPqu6E+C64Yfv1cQ7kz7rIZviUmN+EgEM=
```

Sprig should not appear as an indirect dependency on the `go.mod` file but as a direct one.

LGTM otherwise.

PR #485, opened 2023-01-14
feat: New `eval` function to evaluate a template

rhansen:
This makes it possible to:
  * modularize a template by using an embedded template (defined with Go's built-in `define` action) as a function whose return value is the expansion of the template
  * post-process the expansion of a template (e.g., pipe it to Sprig's `indent` function)

For additional context, see <https://github.com/golang/go/issues/54748>.  Sprig is unlikely to add an equivalent to this function any time soon: the function must be a closure around the template, meaning either Sprig or Go would have to change its API.


rhansen:
Make it easier to modularize.  Technically it's currently possible to use templates as functions, but the return value has to be stored in the template's dot dict instead of the more natural/readable approach of writing it out.  (See https://github.com/nginx-proxy/nginx-proxy/pull/1938 for an example.)

buchdag (approved):

PR #502, opened 2023-03-02
fix: Update list of watchers serially to avoid a data race

rhansen:
Also:
  * Fix some races in tests.
  * Pass `-race` to `go test` to detect data races


internal/generator/generator_test.go
```diff
@@ -189,12 +189,12 @@ func TestGenerateFromEvents(t *testing.T) {
 
 	// The counter is incremented in each output file in the following sequence:
 	//
-	//       init   0ms    5ms    10ms   15ms   20ms   25ms   30ms   35ms   40ms   45ms   50ms   55ms
+	//       init   150ms  200ms  250ms  300ms  350ms  400ms  450ms  500ms  550ms  600ms  650ms  700ms
```
buchdag, line 192: Why does the graph now starts at 150 ms instead of 0 ?

buchdag (commented):
@rhansen I just have one question, other than that LGTM

internal/generator/generator_test.go
rhansen: The original timeline was incorrect—it should have started at 5ms.  (Note that [`eventsResponse` on line 27](https://github.com/nginx-proxy/docker-gen/blob/aafbf65f1a0bbfec2d3b82f61abdfb138a77d234/internal/generator/generator_test.go#L27) starts with a blank line.)

buchdag (approved):

PR #503, opened 2023-03-03
fix: Don't attempt to catch SIGQUIT or SIGKILL

rhansen:
SIGKILL can't be caught so there's no point in trying.  SIGQUIT is intended as a harsh exit for debugging purposes; it should generate a core dump that accurately depicts the program's state at the time the SIGQUIT was received.


rhansen:
Unit tests did not run due to a GitHub outage.  I'll re-trigger them some time after the outage is over.

buchdag (dismissed):

buchdag (approved):

PR #515, opened 2023-04-05
added support to use label for notify-sigup instead of container name

DominikUrban:
This implementation will allow for the use of a filter to live query the containers before sending the sigup, and is fully backwards compatible with the previous implementation.

Resolves https://github.com/nginx-proxy/docker-gen/issues/77

New version of this pull request: https://github.com/nginx-proxy/docker-gen/pull/311


buchdag:
@DominikUrban any particular reason for closing the PR ? 🤔

PR #530, opened 2023-06-01
Emphazise the need for properly setup environment variables

oliv3r:
Some templates rely on environment variables to work properly. This is not really touched on in the documentation. Lets make this a little bit more obvious.


buchdag (changes requested):

buchdag:
Thanks for the PR, aside from a typo in an env variable's name, everything looks good 👍

PR #531, opened 2023-06-03
docker: Add proper entrypoint

oliv3r:
As per docker guidelines [0] a container should always really have a consistent entrypoint, without having to override it or do special tricks.

The behavior should be _identical_ as before, but will no longer trigger errors because docker-gen doesn't understand certain parameters (/bin/sh for example being common). Further more, allows a proper entrypoint for a CI to work easily with the container as well. Allowing for scenario's such as `apk add git && docker-gen renew` in a docker-gen image for example.

E.g. `docker run docker-gen --help` works, as does `docker run docker-gen /bin/sh` or `docker run docker-gen ls`.

[0]: https://github.com/docker-library/official-images#consistency

### DIFF
--- a/app/container-entrypoint.sh
+++ b/app/container-entrypoint.sh
@@ -1,0 +1,17 @@
+#!/bin/sh
+
+set -eu
+
+bin='docker-gen'
+
+# run command if it is not starting with a "-" and is an executable in PATH
+if [ "${#}" -le 0 ] || \
+   [ "${1#-}" != "${1}" ] || \
+   [ -d "${1}" ] || \
+   ! command -v "${1}" > '/dev/null' 2>&1; then
+	entrypoint='true'
+fi
+
+exec ${entrypoint:+${bin:?}} "${@}"
+
+exit 0
--- a/Dockerfile.alpine
+++ b/Dockerfile.alpine
@@ -28,4 +28,5 @@ COPY --from=go-builder /build/docker-gen /usr/local/bin/docker-gen
 # Copy the license
 COPY LICENSE /usr/local/share/doc/docker-gen/
 
-ENTRYPOINT ["/usr/local/bin/docker-gen"]
+COPY "./container-entrypoint.sh" "/app/container-entrypoint.sh"
+ENTRYPOINT [ "/app/container-entrypoint.sh" ]
--- a/Dockerfile.debian
+++ b/Dockerfile.debian
@@ -32,4 +32,5 @@ COPY --from=go-builder /build/docker-gen /usr/local/bin/docker-gen
 # Copy the license
 COPY LICENSE /usr/local/share/doc/docker-gen/
 
-ENTRYPOINT ["/usr/local/bin/docker-gen"]
+COPY "./container-entrypoint.sh" "/app/container-entrypoint.sh"
+ENTRYPOINT [ "/app/container-entrypoint.sh" ]


container-entrypoint.sh
```diff
@@ -0,0 +1,24 @@
...
+
+# run command if it is not starting with a "-" and is an executable in PATH
+if [ "${#}" -le 0 ] || \
+   [ "${1#-}" != "${1}" ] || \
+   [ -d "${1}" ] || \
+   ! command -v "${1}" > '/dev/null' 2>&1; then
+	entrypoint='true'
+fi
+
+exec ${entrypoint:+${bin:?}} "${@}"
```
buchdag, line 22: What's the rationale behind the `[ -d "${1}" ]` check (this would end up passing a folder as argument to docker-gen, and AFAIK this does not work) ?  🤔

While I understand the `command -v "${1}" > '/dev/null' 2>&1` check, I don't think that's the way to go because it would mean commands with typo would be passed as argument to docker-gen instead of failing with a `foo: not found` as they should. A simple `[ -f "${1}" ]` work in this case: executables in `PATH` do not check as files.

Finally the `exec ${entrypoint:+${bin:?}} "${@}"` is hard to decipher and needlessly complex, the way it's done in the official docker example is clearer.

Suggestion taking all of the above into account:

```shell
#!/bin/sh

set -eu

# if command starts with an "-" or is a file, pass it as argument(s) to docker-gen, else run it.
if [ "${#}" -le 0 ] || [ "${1#-}" != "${1}" ] || [ -f "${1}" ]; then
	set -- docker-gen "${@}"
fi

exec "${@}"
```

The `exit 0` at the end as no real use since we're using `set -e`.

buchdag (changes requested):
Hi. Thanks for the PR.

In addition to the change requests I made below, I'd prefer the entrypoint script to be located at `/app/docker-entrypoint.sh`, both in the repo and inside the image, for consistency with https://github.com/nginx-proxy/nginx-proxy

container-entrypoint.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/sh
+# SPDX-License-Identifier: MIT
+#
+# Copyright (C) 2023 Olliver Schinagl <oliver@schinagl.nl>
+#
+# A beginning user should be able to docker run image bash (or sh) without
+# needing to learn about --entrypoint
+# https://github.com/docker-library/official-images#consistency
```
buchdag, line 8: Missed that: please do not commit personal copyright informations and/or license informations to open source projects. Also the whole commented out section isn't necessary IMHO, the rationale for adding a proper entrypoint was already clear enough in the commit comments.

oliv3r: You know that even in the linux kernel this is very normal, it's the 'standard' 'author' if you will line. Even if you do a CLA, you sign over your copyright, but still retain your own. Heck, this even went through cisco's legal department, where the only request was, to also add their copyright (as per CLA).

I'm fine with removing it, I don't care, but this is very standard 'run-of-the-mill' stuff, done pretty much in every opensource project ;)

As for the 'explanation' sure, I don't mind removing it :)

container-entrypoint.sh
oliv3r: It looks like github ate my comment, as I'm very sure I replied in detail. I'll try to find it again.

container-entrypoint.sh
buchdag: No, I did not know, I've never seen it done even once on any project I helped maintain or contributed to over the years (but admittedly they're not on the same level as the linux kernel or Cisco stuff), so I assumed something that might have been wrong.

I don't have strong feeling about it aside from finding it somewhat odd to do it on a repo where this was never ever done by any other contributor. If you rather commit your copyright go ahead.

oliv3r: I don't mind, I've used this same template for years now; But it's your repo, your the maintainer, so I'll do what you prefer, well already did :)

container-entrypoint.sh
buchdag: Aw crap. 😞

oliv3r: Yep, gone; was hoping I could get it in the edit field on my laptop, no such luck ... so let me try again.

> What's the rationale behind the `[ -d "${1}" ]` check (this would end up passing a folder as argument to docker-gen, and AFAIK this does not work) ? thinking

I have to dig deep now, as I use this template for everything (to keep them all the same, nice and consistent). Afaik, if you have a container, that accepts a directory as input, lets say the `ls` container, if the directory exists, it'll allow you to pass the directory as argument to the binary. But you got me, I'm trying remember really hard what the usecase was (there was one :p)

> 
> While I understand the `command -v "${1}" > '/dev/null' 2>&1` check, I don't think that's the way to go because it would mean commands with typo would be passed as argument to docker-gen instead of failing with a `foo: not found` as they should. A simple `[ -f "${1}" ]` work in this case: executables in `PATH` do not check as files.

But was it a command you where trying to execute? or was it something you wanted to pass a long. If you assume the script to be dedicated only to `docker-gen` yeah, sure. But in a more generic case. Also, if I supply `cp` instead of `/usr/bin/cp` it should still work.

> 
> Finally the `exec ${entrypoint:+${bin:?}} "${@}"` is hard to decipher and needlessly complex, the way it's done in the official docker example is clearer.

Eye of the beholder, sure enough. I used to have `${bin:+${bin}}` before, is that easier on the eyes? I took this new approach, as I wanted to keep the bin definition at the top of the file, for easy recognition.

The docker approach is equally 'confusing'. What's with the `set --`, that is also less clear (doesnt' that eat the existing `set -eu` also? Idk, never had to use it yet). The reason it is a variable though; is to ensure a single exit point of the script.

> 
> Suggestion taking all of the above into account:
> 
> ```shell
> #!/bin/sh
> 
> set -eu
> 
> # if command starts with an "-" or is a file, pass it as argument(s) to docker-gen, else run it.
> if [ "${#}" -le 0 ] || [ "${1#-}" != "${1}" ] || [ -f "${1}" ]; then
>     set -- docker-gen "${@}"
> fi
> 
> exec "${@}"
> ```
> 
> The `exit 0` at the end as no real use since we're using `set -e`.
ah, but what if some hacker dude does `echo 'rm -rf /' >> script.sh`? It's a shell scripting best-practise (afaik its in the google bash styleguide even). I like best-practises :) 'always have a defined exit point at the end of your script', in case something goes haywire.

app/container-entrypoint.sh
```diff
@@ -0,0 +1,17 @@
...
+
+# run command if it is not starting with a "-" and is an executable in PATH
+if [ "${#}" -le 0 ] || \
+   [ "${1#-}" != "${1}" ] || \
+   [ -d "${1}" ] || \
+   ! command -v "${1}" > '/dev/null' 2>&1; then
+	entrypoint='true'
+fi
+
+exec ${entrypoint:+${bin:?}} "${@}"
```
rhansen, line 15: `entrypoint` should be initialized to the empty string in case it is already exported to this script's environment.

app/container-entrypoint.sh
```diff
@@ -0,0 +1,17 @@
...
+
+exit 0
```
rhansen, line 17: Not needed; `exec` cannot return unless it can't execute the command, in which case `exec`'s return value should be returned.

```suggestion
```

app/container-entrypoint.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
+
+set -eu
+
+bin='docker-gen'
+
+# run command if it is not starting with a "-" and is an executable in PATH
+if [ "${#}" -le 0 ] || \
+   [ "${1#-}" != "${1}" ] || \
+   [ -d "${1}" ] || \
```
rhansen, line 10: Isn't this implied by the `command` check?

app/container-entrypoint.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
+
+set -eu
+
+bin='docker-gen'
+
+# run command if it is not starting with a "-" and is an executable in PATH
+if [ "${#}" -le 0 ] || \
+   [ "${1#-}" != "${1}" ] || \
```
rhansen, line 9: If you pass `--` to `command`, it will prevent `$1` from being interpreted as an option. That would make it possible to eliminate this check. (I don't think it's a problem accidentally running a command named something like `-x`; such commands shouldn't exist, and if they do then running it is no worse than accidentally running any other command that was intended to be a command-line option to `docker-gen`.)

app/container-entrypoint.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
+
+set -eu
+
+bin='docker-gen'
+
+# run command if it is not starting with a "-" and is an executable in PATH
+if [ "${#}" -le 0 ] || \
```
rhansen, line 8: `command` will return 1 if the command name is the empty string, so this check isn't necessary.

app/container-entrypoint.sh
rhansen, line 15: I agree with @buchdag that this should use `set -- docker-gen "$@"`.  It is more idiomatic and thus more readable.

app/container-entrypoint.sh
rhansen, line 17: Putting it all together, I would do something like this:

```suggestion
command -v -- "$1" >/dev/null 2>&1 || set -- docker-gen "$@"
exec "$@"
```

rhansen (changes requested):

buchdag:
Superseded by #573

app/container-entrypoint.sh
oliv3r: Thank you for this. I've gone over this a lot in my mind over the year. Could you be right on all cases. Could my approach have been improved? I think the answer is yes, and you showed me that. I've adapted it for my general purpose needs as such:

```
# Prefix args with $bin if $1 is not a valid command
if ! command -v -- "${1:-}" > '/dev/null' 2>&1; then
	set -- "${bin:?}" "${@}"
fi
exec "${@}"

exit 0
```

I think it's shorter and more concise. I did use an 'if/else' structure instead of an 'or' structure as I find it's easier to read for passers by.

I use 'bin' instead of just pasting the command, as I reuse this snippet in multiple entry-point scripts, and having it re-usable is valuable :)

Also I'm keeping the exit 0 at the end; it's a security thing and should be done consistently. Not 'oh maybe I don't need it here', also, again re-usability.

Also, style-wise I follow the common best practice to always use curly-braces around variables.

But even so, thank you very much for this improvement! You helped me quite a bit to make my standard entrypoint even more ultimate :)

PR #548, opened 2023-08-26
Add support to pull data from multiple docker endpoints (e.g. swarm)

tarasov65536:
This PR adds support to request container data from multiple docker endpoints. For example in swarm you now can get containers information from all nodes at once.

Main idea is that `endpoint` parameter refers to local docker node for which docker-gen generates configs. And the `swarm-node` parameters define additional docker nodes from which docker-gen pulls containers information.

Back compatability preserved. If no `swarm-node` stated then docker-gen collects information only from local docker endpoint defined by `endpoint` parameter.

In case when stated at least one `swarm-node` parameter docker-gen will collect containers information from swarm-node list. And use `endpoint` only for "controlling" purpose (e.g. to notify containers).

Root use-case for this PR is to make possible for projects like [nginx-proxy](https://github.com/nginx-proxy/nginx-proxy) to have full information across all swarm cluster.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,6 +96,8 @@ Options:
       config files with template directives. Config files will be merged if this option is specified multiple times. (default [])
   -endpoint string
       docker api endpoint (tcp|unix://..). Default unix:///var/run/docker.sock
+  -swarm-node value
+      docker api endpoints from which to listen for events. Default equals to value of `endpoint` argument
   -interval int
       notify command interval (secs)
   -keep-blank-lines
--- a/cmd/docker-gen/main.go
+++ b/cmd/docker-gen/main.go
@@ -34,6 +34,7 @@ var (
 	interval              int
 	keepBlankLines        bool
 	endpoint              string
+	swarmNodes            stringslice
 	tlsCert               string
 	tlsKey                string
 	tlsCaCert             string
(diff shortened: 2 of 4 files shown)


tarasov65536:
Hi @buchdag ,

Thanks for your reply!

I’ve just merged the latest `main` branch into the PR and resolved all conflicts. The branch is now up to date and ready for review.

Regarding your question: the core idea of Swarm support in this PR is that we can define multiple `swarm-node` endpoints from which to collect container metadata (via the Docker API). However, actions like `sendSignalToContainer`, `SetServerInfo`, and `SetDockerEnv` are only ever performed against the single `endpoint`, which represents the **local Docker engine** — the one where docker-gen is running.

So, even in Swarm mode:
* All container data is aggregated from multiple remote Docker nodes (via `swarm-node`).
* But any control actions (e.g., sending signals, updating env) are only done against the local `endpoint`.

This matches the typical use case I have in production:
I run a Docker Swarm with multiple nodes, and some of them act as **gateways**. On each of these gateway nodes, I run a pair of containers: `docker-gen` and `nginx`.
Each `docker-gen` instance collects container info from all Swarm nodes using the `swarm-node` parameters and then generates the **local nginx** config accordingly — and only restarts its local nginx instance.

This setup works very reliably for me, and I’m hoping that this PR can help push that logic further into the `nginx-proxy` project as well — eventually allowing a single container (with docker-gen logic embedded) to handle everything instead of needing two containers.

PR #549, opened 2023-08-26
Filter containers to notify

tarasov65536:
This PR gives ability to specify set of filters to containers query when new generated file arrived.

It is kinda similar to [#311](https://github.com/nginx-proxy/docker-gen/pull/311) but with better implementation from architectural point of view.

It introduces new command line argument `-notify-filter` that adds filter in list of filters to query containers that should be notified when generated file changed. Argument format as in [docker api](https://docs.docker.com/engine/reference/commandline/ps/#filter)

For example:
```sh
docker-gen -notify-filter name=foo -notify-filter label=bar=baz ...
```
adds two filters to containers query.

Signal code still defined by argument `-notify-signal`.


josegonzalez:
This probably needs a test in order to get merged.

tarasov65536:
@josegonzalez, what kind of test it needs? Can't find anything similar.

buchdag:
Superseded by #619

PR #558, opened 2023-09-30
feat: add `Internal` property to `Network` struct

pini-gh:
The purpose of this pull request is to add the `Internal` property to the `Network` structure so that it could be used to fix https://github.com/nginx-proxy/nginx-proxy/issues/2310.

I've tested that this change makes the fix for the nginx-proxy related issue easy:
```
diff --git a/nginx.tmpl b/nginx.tmpl
index fb0766b..afe5c96 100644
--- a/nginx.tmpl
+++ b/nginx.tmpl
@@ -75,7 +75,7 @@
                 {{- continue }}
             {{- end }}
             {{- range sortObjectsByKeysAsc $.globals.CurrentContainer.Networks "Name" }}
-                {{- if and . .Gateway }}
+                {{- if and . .Gateway (not .Internal) }}
     #         container is in host network mode, using {{ .Name }} gateway IP
                     {{- $ip = .Gateway }}
                     {{- break }}
```

I have no knowledge in Go programming and it took me quite some time to come up with this proposal. Please bear with me :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -258,6 +258,7 @@ type Network struct {
     MacAddress          string
     GlobalIPv6PrefixLen int
     IPPrefixLen         int
+    Internal            bool
 }
 
 type DockerImage struct {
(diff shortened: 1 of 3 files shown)


buchdag (approved):

buchdag:
Nice !

PR #607, opened 2024-05-05
feat: add WithDefault variants to groupBy template functions

buchdag:
The PR add two new template function:
- `groupByWithDefault $containers $fieldPath $defaultValue`
- `groupByLabelWithDefault $containers $label $defaultValue`.

They both work as their counterpart `groupBy` and `groupByLabel`, but do not omit the containers that do not have a value for the provided field path expression or label, and instead group those under the provided `$defaultValue`.

I'm less convinced of the usefulness of a `groupByMultiWithDefault` template function, but maybe I'm wrong.

I'm not certain how a `groupByKeysWithDefault $containers $fieldPath $defaultValue` should work to be most useful and consistent:
- return an array containing only the `$defaultValue` when **no containers** have the field path set ?
- return an array containing the `$defaultValue` for **every container** that don't have the field path set ?

ping @pini-gh


pini-gh:
> I'm less convinced of the usefulness of a `groupByMultiWithDefault` template function, but maybe I'm wrong.

It's not needed for nginx-proxy, but could be for other use cases.

> I'm not certain how a `groupByKeysWithDefault $containers $fieldPath $defaultValue` should work to be most useful and consistent:

* return an array containing the `$defaultValue` when **at least one container** doesn't have the field path set

buchdag:
This solution and the second one I proposed might cause issues in cases where we do stuff like this:
```
{{- $foot := or (first (groupByKeys $containers "Env.FOO")) "" }}
```
If we replace it with:
```
{{- $foot := first (groupByKeysWithDefault $containers "Env.FOO" "") }}
```
If the first container iterated over doesn't have the field path set, `$foo` is assigned the default value despite other containers having a value set for `Env.FOO`

pini-gh:
> If the first container iterated over doesn't have the field path set, `$foo` is assigned the default value despite other containers having a value set for `Env.FOO`

Yes, indeed. But it makes no sense to use `groupByKeysWithDefault` when the purpose is to get the first non-default value.

buchdag:
If that's okay with you I'll merge this PR as is and release a new version of docker-gen, then come back later for `groupByMultiWithDefault` and `groupByKeysWithDefault`.

In the context of nginx-proxy's template, I'm not really convinced `groupByLabelWithDefault` will immediately help us either because we're using `groupByLabel` kind of for the same purpose as `groupByKeys` (get the first set value of a given label or field path among a list of containers).

pini-gh:
Yes that's fine.

PR #713, opened 2025-11-26
feat: Adding new Notify Option using exec

Ozymandias88:
New NotifyContainersCmd option that uses exec to run a command on a container when a file is generated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,9 +223,18 @@ dest = "/etc/nginx/conf.d/default.conf"
 watch = true
 wait = "500ms:2s"
 
+[[config]]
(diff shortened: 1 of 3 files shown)


JamBalaya56562:
This looks like a useful addition — `NotifyContainers` can only send signals, so a
container that reloads via a command (your `pihole reloaddns` case) has no way to be
notified today.

Two things before it can move forward:

- The branch now conflicts with `main`. `internal/config/config.go` and
  `internal/generator/generator.go` have both changed since November, so it needs a rebase.
- Could you add test coverage for `sendCmdToContainers`?

Are you still planning to work on this?

Issue #238, opened 2017-02-17
TestGenerateFromEvents race condition


adrianlzt:
Several executions of the ``generator_test.go`` returns different outputs (I have deleted the rest of _test.go files).

```
$ go version  
go version go1.7.5 linux/amd64

$ git log -1
commit cb6223c73c82c4fbecae3a8d972d856ed6f026db
```
```
$ go test -v                 
=== RUN   TestGenerateFromEvents
--- FAIL: TestGenerateFromEvents (0.58s)
	generator_test.go:204: expected: 8dfafdbc3a40.5. got: 8dfafdbc3a40.6
	generator_test.go:204: expected: 8dfafdbc3a40.6. got: 8dfafdbc3a40.7
	generator_test.go:204: expected: 8dfafdbc3a40.7. got: 8dfafdbc3a40.5
FAIL
exit status 1
FAIL	_/var/tmp/docker-gen	0.584s

$ go test -v
=== RUN   TestGenerateFromEvents
--- PASS: TestGenerateFromEvents (0.58s)
PASS
ok  	_/var/tmp/docker-gen	0.585s

$ go test -v
=== RUN   TestGenerateFromEvents
--- FAIL: TestGenerateFromEvents (0.60s)
	generator_test.go:204: expected: 8dfafdbc3a40.6. got: 8dfafdbc3a40.7
	generator_test.go:204: expected: 8dfafdbc3a40.7. got: 8dfafdbc3a40.6
FAIL
exit status 1
FAIL	_/var/tmp/docker-gen	0.601s

$ go test -v          
=== RUN   TestGenerateFromEvents
--- PASS: TestGenerateFromEvents (0.58s)
PASS
ok  	_/var/tmp/docker-gen	0.585s

$ go test -v
=== RUN   TestGenerateFromEvents
--- PASS: TestGenerateFromEvents (0.58s)
PASS
ok  	_/var/tmp/docker-gen	0.585s

$ go test -v
=== RUN   TestGenerateFromEvents
--- PASS: TestGenerateFromEvents (0.58s)
PASS
ok  	_/var/tmp/docker-gen	0.585s

$ go test -v
=== RUN   TestGenerateFromEvents
--- FAIL: TestGenerateFromEvents (0.58s)
	generator_test.go:204: expected: 8dfafdbc3a40.6. got: 8dfafdbc3a40.7
	generator_test.go:204: expected: 8dfafdbc3a40.7. got: 8dfafdbc3a40.6
FAIL
[...]

PR #744, opened 2026-06-18
fix: emit Networks and Addresses in deterministic order

### LINKED ISSUE #169: Support for multiple docker networks (via libnetwork)
Hi there, first of all, thanks a lot for this project, has worked like a charm so far.
My issue is the following : since I separated my containers in multiples networks in docker, I can't seem to produce the right template to proxy to a container belonging to 2 networks properly, it always gets the wrong IP address (the one on the backend network, not on the frontend network). My config : 
- backend_network:
  - gitlab
  - postgresql
  - redis
- frontend_network
  - nginx
- other_network
  - docker-gen
  - letsencrypt automation

I'm using the following template to generate the nginx config file :
```nginx
server {
    listen                      80 default_server;
    listen                      [::]:80 default_server;
    server_name         _;

    # Letsencrypt challenge handling
    location /.well-known/acme-challenge/ {
        root                     /usr/share/nginx/html;
        try_files               $uri =404;
    }

    # Default behaviour for non ssl/tls requests
    location / {
        return                  301 https://$host$request_uri;
    }
}

{{ range $host, $containers := groupByMulti $ "Env.VIRTUAL_HOST" "," }}
upstream {{ $host }} {

{{ range $index, $value := $containers }}

    {{ $addrLen := len $value.Addresses }}
    {{ $network := index $value.Networks 0 }}

    {{/* If only 1 port exposed, use that */}}
    {{ if eq $addrLen 1 }}
        {{ with $address := index $value.Addresses 0 }}
            # {{$value.Name}}
            server {{ $network.IP }}:{{ $address.Port }};
        {{ end }}

    {{/* If more than one port exposed, use the one matching VIRTUAL_PORT env var */}}
    {{ else if $value.Env.VIRTUAL_PORT }}
        {{ range $i, $address := $value.Addresses }}
            {{ if eq $address.Port $value.Env.VIRTUAL_PORT }}
            # {{$value.Name}}
            server {{ $network.IP }}:{{ $address.Port }};
            {{ end }}
        {{ end }}

    {{/* Else default to standard web port 80 */}}
    {{ else }}
        {{ range $i, $address := $value.Addresses }}
            {{ if eq $address.Port "80" }}
            # {{$value.Name}}
            server {{ $network.IP }}:{{ $address.Port }};
            {{ end }}
        {{ end }}
    {{ end }}
{{ end }}
}
server {
    server_name                 {{ $host }};
    listen                      443 ssl;
    listen                      [::]:443 ssl;

    ssl_certificate             /etc/nginx/certs/{{ trim $host }}.crt;
    ssl_certificate_key         /etc/nginx/certs/{{ trim $host }}.key;
    ssl_dhparam                 /etc/nginx/certs/dhparam.pem;

    ssl_session_timeout         1d;
    ssl_session_cache           shared:SSL:50m;
    ssl_protocols               TLSv1 TLSv1.1 TLSv1.2;
    ssl_ciphers                 HIGH:!aNULL:!MD5;
    ssl_prefer_server_ciphers   on;

    error_log                   /proc/self/fd/2;
    access_log                  /proc/self/fd/1;

    location / {
        gzip                            off;
        proxy_buffering                 off;
        proxy_set_header X-Forwarded-Ssl on;

        proxy_pass http://{{ trim $host }};
        proxy_set_header Upgrade $http_upgrade;
        proxy_set_header Connection "upgrade";
        proxy_set_header Host $http_host;
        proxy_set_header X-Real-IP $remote_addr;
        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
        proxy_set_header X-Forwarded-Proto $scheme;
        proxy_set_header X-Frame-Options SAMEORIGIN;
    }

    # 6G Perishable Press: Queries String
    # @ https://perishablepress.com/6g/
    location ~* "(eval\()"  { return 444; }
    location ~* "(127\.0\.0\.1)"  { return 444; }
    location ~* "([a-z0-9]{2000})"  { return 444; }
    location ~* "(javascript\:)(.*)(\;)"  { return 444; }
    location ~* "(base64_encode)(.*)(\()"  { return 444; }
    location ~* "(GLOBALS|REQUEST)(=|\[|%)"  { return 444; }
    location ~* "(<|%3C).*script.*(>|%3)" { return 444; }
    location ~ "(\\|\.\.\.|\.\./|~|`|<|>|\|)" { return 444; }
    location ~* "(boot\.ini|etc/passwd|self/environ)" { return 444; }
    location ~* "(thumbs?(_editor|open)?|tim(thumb)?)\.php" { return 444; }
    location ~* "(\'|\")(.*)(drop|insert|md5|select|union)" { return 444; }

    # 6G Perishable Press : Request String
    # @ https://perishablepress.com/6g/
    location ~* "(https?|ftp|php):/" { return 444; }
    location ~* "(=\\\'|=\\%27|/\\\'/?)\." { return 444; }
    location ~* "/(\$(\&)?|\*|\"|\.|,|&|&amp;?)/?$" { return 444; }
    location ~ "(\{0\}|\(/\(|\.\.\.|\+\+\+|\\\"\\\")" { return 444; }
    location ~ "(~|`|<|>|:|;|,|%|\\|\s|\{|\}|\[|\]|\|)" { return 444; }
    location ~* "/(=|\$&|_mm|(wp-)?config\.|cgi-|etc/passwd|muieblack)" { return 444; }
    location ~* "(&pws=0|_vti_|\(null\)|\{\$itemURL\}|echo(.*)kae|etc/passwd|eval\(|self/environ)" { return 444; }
    location ~* "\.(aspx?|bash|bak?|cfg|cgi|dll|exe|git|hg|ini|jsp|log|mdb|out|sql|svn|swp|tar|rar|rdf)$" { return 444; }
    location ~* "/(^$|mobiquo|phpinfo|shell|sqlpatch|thumb|thumb_editor|thumbopen|timthumb|webshell)\.php" { return 444; }
}
{{ end }}
```

Any thoughts, or is the functionnality yet to be implemented ?

### LINKED ISSUE #431: order of networks in nginx template is random
I'm not quite sure if this is a `docker-gen` issue or maybe a `nginx-proxy` template issue.

I'm using the three container setup for a nginx proxy and noticed multiple `nginx` restarts due to changed `default.conf`

Comparing the generated conf files showed me, that the changes are only for the network comment lines.

```
/nginx-proxy # diff -W 90 -y --suppress-common-lines generated-11.conf generated-12.conf
        # Cannot connect to network 'elm    <
					    >	        # Cannot connect to network 'elmar
        # Cannot connect to network 'git    <
					    >	        # Cannot connect to network 'gitla
        # Cannot connect to network 'hc-    <
					    >	        # Cannot connect to network 'hc-vo
        # Cannot connect to network 'sec    <
					    >	        # Cannot connect to network 'secon
					    >	        # Cannot connect to network 'stats
        # Cannot connect to network 'sta    <
        # Cannot connect to network 'elm    <
					    >	        # Cannot connect to network 'elmar
        # Cannot connect to network 'sec    <
					    >	        # Cannot connect to network 'secon
```

Related "discussion": https://github.com/nginx-proxy/nginx-proxy/discussions/1981#discussioncomment-2667448

JamBalaya56562:
## Summary

The per-container `Networks` and `Addresses` slices exposed to templates are built by ranging over Go maps (`NetworkSettings.Networks`, `NetworkSettings.Ports` and `Config.ExposedPorts`), whose iteration order is random. As a result:

- Generated files (e.g. an nginx `default.conf`) churn on every regeneration even when nothing meaningful changed, triggering spurious reloads / container restarts (#431).
- `index .Networks 0` picks an arbitrary network, so multi-network IP selection is unstable (#169).

This PR sorts both slices into a deterministic, total order so the rendered output is stable across regenerations. Routing is unaffected — only the ordering becomes deterministic.

## Changes

- `internal/context/address.go`: sort addresses in `GetContainerAddresses` via a new `sortAddresses` helper — by port (compared **numerically**, so `80 < 443 < 8080`), then proto, host port, host IP, IP.
- `internal/generator/generator.go`: sort `runtimeContainer.Networks` by `Name` (the unique network-name map key) via a new `sortNetworks` helper, right after the network map is collected.
- Tests: add order-sensitive `TestSortAddresses`, `TestGetContainerAddressesSorted` and `TestSortNetworks`. The existing order-insensitive address tests are unchanged and still pass.
- `README.md`: document the deterministic ordering and how to select a network by name.

All new symbols are unexported and the change is purely additive.

## Behavior note

`index .Networks 0` now returns the alphabetically-first network (e.g. `bridge`) instead of an arbitrary one — this is the intended, deterministic fix (as suggested in #431). A specific network can still be selected by name:

```
{{ $net := index (where $value.Networks "Name" "my-network") 0 }}
```

## Resolves

- Fixes #431
- Fixes #169 — no further code change is needed; a specific network is selectable via `where` once the ordering is stable.

---

_Note: an unrelated, pre-existing flake in `TestGenerateFromEvents` (timing-dependent; see #238) can fail on the Windows runner — it reproduces on `main` without this change (verified locally on Windows / Go 1.26). It is independent of this PR._

### DIFF
--- a/README.md
+++ b/README.md
@@ -394,6 +394,20 @@ For example, this is a JSON version of an emitted RuntimeContainer struct:
 }
 ```
 
+#### Ordering of `Addresses` and `Networks`
+
+The `Addresses` and `Networks` slices are emitted in a deterministic order, so generated output is stable across regenerations and does not trigger spurious reloads/restarts when nothing meaningful changed:
+
+- `Addresses` are sorted by port (compared numerically), then by protocol, host port, host IP and IP.
+- `Networks` are sorted alphabetically by `Name`.
+
+As a consequence, `index .Networks 0` returns the alphabetically-first network (for example `bridge`) rather than an arbitrary one. To select a specific network by name, filter with `where`:
+
+```
+{{ $net := index (where $value.Networks "Name" "my-network") 0 }}
+server {{ $net.IP }}:{{ (index $value.Addresses 0).Port }};
+```
+
 #### Functions
 
 - [Functions from Go](https://pkg.go.dev/text/template#hdr-Functions)
--- a/internal/context/address.go
+++ b/internal/context/address.go
@@ -1,6 +1,9 @@
 package context
 
 import (
+	"sort"
+	"strconv"
+
 	docker "github.com/fsouza/go-dockerclient"
 )
 
@@ -46,5 +49,29 @@ func GetContainerAddresses(container *docker.Container) []Address {
 		}
 	}
 
+	sortAddresses(addresses)
+
 	return addresses
+}
+
+// sortAddresses sorts addresses in place by port (numeric), then proto, host port, host IP and IP.
+func sortAddresses(addresses []Address) {
+	sort.Slice(addresses, func(i, j int) bool {
+		a, b := addresses[i], addresses[j]
+		pa, _ := strconv.Atoi(a.Port)
+		pb, _ := strconv.Atoi(b.Port)
+		if pa != pb {
+			return pa < pb
+		}
+		if a.Proto != b.Proto {
+			return a.Proto < b.Proto
+		}
+		if a.HostPort != b.HostPort {
+			return a.HostPort < b.HostPort
+		}
+		if a.HostIP != b.HostIP {
+			return a.HostIP < b.HostIP
+		}
+		return a.IP < b.IP
+	})
 }
--- a/internal/context/address_test.go
+++ b/internal/context/address_test.go
@@ -114,3 +114,50 @@ func TestGenerateContainerAddressesWithNoPorts(t *testing.T) {
 		HostPort:     "",
 	})
 }
+
+func TestSortAddresses(t *testing.T) {
+	addresses := []Address{
+		{IP: "10.0.0.10", Port: "8080", Proto: "tcp"},
+		{IP: "10.0.0.10", Port: "80", Proto: "tcp"},
+		{IP: "10.0.0.10", Port: "443", Proto: "tcp"},
+		{IP: "10.0.0.10", Port: "53", Proto: "udp"},
+		{IP: "10.0.0.10", Port: "53", Proto: "tcp"},
+	}
+
+	// Port sorts numerically (not lexically: "443" must not sort before "80"),
+	// with Proto as the tie-breaker for equal ports (53/tcp before 53/udp).
+	want := []Address{
+		{IP: "10.0.0.10", Port: "53", Proto: "tcp"},
+		{IP: "10.0.0.10", Port: "53", Proto: "udp"},
+		{IP: "10.0.0.10", Port: "80", Proto: "tcp"},
+		{IP: "10.0.0.10", Port: "443", Proto: "tcp"},
+		{IP: "10.0.0.10", Port: "8080", Proto: "tcp"},
+	}
+
+	sortAddresses(addresses)
+	assert.Equal(t, want, addresses)
+}
+
+func TestGetContainerAddressesSorted(t *testing.T) {
+	testContainer := &docker.Container{
+		Config: &docker.Config{
+			ExposedPorts: map[docker.Port]struct{}{},
+		},
+		NetworkSettings: &docker.NetworkSettings{
+			IPAddress: "10.0.0.10",
+			Ports:     map[docker.Port][]docker.PortBinding{},
+		},
+	}
+	// Insert ports out of numeric order; the map iteration order is random but
+	// GetContainerAddresses must return them deterministically sorted by port.
+	testContainer.NetworkSettings.Ports[httpTestPort] = []docker.PortBinding{} // 8080
+	testContainer.NetworkSettings.Ports[httpPort] = []docker.PortBinding{}     // 80
+	testContainer.NetworkSettings.Ports[httpsPort] = []docker.PortBinding{}    // 443
+
+	addresses := GetContainerAddresses(testContainer)
+	ports := make([]string, len(addresses))
+	for i, a := range addresses {
+		ports[i] = a.Port
+	}
+	assert.Equal(t, []string{"80", "443", "8080"}, ports)
+}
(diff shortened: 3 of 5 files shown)


buchdag (approved):

buchdag:
@JamBalaya56562 please do not include issue or PR numbers in either PR titles or commit message.

buchdag (commented):
> sortAddresses sorts addresses in place by port (numeric), then proto, host port, host IP and IP.

I'm not 100% certain about the rationale of sorting by IP last, could you elaborate ? 🤔

Not that I'm against it but I'm not quite sure what sorting strategy would make the most sense for a slice of `Address`.

JamBalaya56562:
Good question. The key thing is that the only fields which actually vary between the addresses of a single container are `Port` and `Proto` — `IP`, `IP6LinkLocal` and `IP6Global` all come straight from the one `container.NetworkSettings.*` in `renderAddress`, so they're identical across every `Address` we emit for that container and can't discriminate between them at all. That's really the answer to "why IP last": it's the weakest possible key here, since it never changes within the slice.

`Port` goes first because it's the field templates actually reason about and index on — the documented common access is `(index .Addresses 0).Port` — and sorting it numerically (via `strconv.Atoi`) gives the intuitive 80 < 443 < 8080 rather than the lexical "443" < "80". `Proto` is the natural next key for the real case where the same port appears as both tcp and udp (e.g. `53/tcp` vs `53/udp`).

`HostPort`, `HostIP` and `IP` are then just there to make the comparator a *total* order. The whole point of the PR is deterministic output, and if two entries ever compared equal on the meaningful keys, `sort.Slice` would leave their relative order unspecified and we'd be back to depending on `Ports`/`ExposedPorts` map iteration order — exactly the nondeterminism we're removing. `IP` lands last because it's the most arbitrary (Docker-assigned, not user-chosen) and, being container-wide, almost never decides anything — it's a final backstop for totality rather than a key we expect to matter. Totally happy to reorder the tie-breakers, or promote a different primary key, if another precedence reads better to you; the determinism holds regardless of how we rank them.

PR #755, opened 2026-07-10
fix: don't treat image namespace as registry

### LINKED ISSUE #154: docker-gen incorrectly parses the user name part of the Docker image as the registry
docker-gen incorrectly parses the user repository part of an image name as the registry name. For example, for an named "tianon/centos", docker-gen will consider that "tianon" is the registry, whereas it's the user repository.

The bug is in the [splitDockerImage](https://github.com/jwilder/docker-gen/blob/cb69dac13db1e7f765730544d01e63396a744252/docker_client.go#L102) function: the current logic is to consider anything before the first '/' as the registry name, which is incorrect.

The correct logic can be seen from the [Docker source code](https://github.com/docker/docker/blob/master/reference/reference.go#L157). If the part before the first '/' has a dot or a colon or if it's equal to "localhost", then it's the registry, otherwise it's the user repository.

JamBalaya56562:
## Problem

`SplitDockerImage` treats everything before the first `/` as the registry. For Docker Hub images that use a namespace/user prefix — e.g. `tianon/centos` — this incorrectly yields `Registry="tianon"`, `Repository="centos"` instead of `Registry=""` (implicit Docker Hub) and `Repository="tianon/centos"`.

Per Docker's reference parser, the first path component is a registry **only** when it contains a `.` or `:`, or is exactly `localhost`. Otherwise it belongs to the repository path.

## Change

`SplitDockerImage` now applies that rule: the first segment is treated as the registry only when it contains `.`/`:` or equals `localhost`; otherwise the whole `user/image` is kept as the repository. Tag parsing is unchanged.

Because `DockerImage.String()` rejoins `registry + "/" + repository`, the full image string still round-trips — only the split `.Image.Registry` / `.Image.Repository` fields change, which is the intended correction for templates that read those fields.

## Tests

Added `tianon/centos` and `tianon/centos:7` cases. All existing cases (`ubuntu`, `custom.registry/…`, `localhost:8888/…`, etc.) still pass. No new dependencies.

Fixes #154

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/internal/dockerclient/docker_cli.go
+++ b/internal/dockerclient/docker_cli.go
@@ -119,20 +119,18 @@ func parseHost(addr string) (string, string, error) {
 }
 
 func SplitDockerImage(img string) (string, string, string) {
-	index := 0
-	repository := img
 	var registry, tag string
-	if strings.Contains(img, "/") {
-		separator := strings.Index(img, "/")
-		registry = img[index:separator]
-		index = separator + 1
-		repository = img[index:]
+	repository := img
+	if i := strings.Index(img, "/"); i != -1 {
+		if first := img[:i]; strings.ContainsAny(first, ".:") || first == "localhost" {
+			registry = first
+			repository = img[i+1:]
+		}
 	}
 
-	if strings.Contains(repository, ":") {
-		separator := strings.Index(repository, ":")
-		tag = repository[separator+1:]
-		repository = repository[0:separator]
+	if i := strings.Index(repository, ":"); i != -1 {
+		tag = repository[i+1:]
+		repository = repository[:i]
 	}
 
 	return registry, repository, tag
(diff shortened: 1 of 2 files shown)


buchdag:
@JamBalaya56562 please flag AI-generated PR explicitly in the PR description (ie "🤖 Generated with [Claude Code](https://claude.com/claude-code)" like in [this PR](https://github.com/nginx-proxy/docker-gen/pull/747)) and add the AI agent as co-author of the commits.

JamBalaya56562:
Done — added the `Co-Authored-By: Claude Opus 4.8` trailer to the commit and the `🤖 Generated with [Claude Code](https://claude.com/claude-code)` note to the PR description. I also rebased onto the latest `main`. Thanks for the review!

buchdag (approved):

PR #771, opened 2026-07-24
feat: basic template validation function

buchdag:
This PR introduce three template functions intended to be used as validators.

All three take a string as last input and either return it unmodified or return an error, so that they can be used in a go template pipeline and prevent docker-gen from rendering the template if their respective validation fail.

- _`mustBeOneOf $slice $value`_: Validates that `$value` is one of the allowed string values in `$slice`.
- _`mustBeInt $value`_: Validates that `$value` is a base-10 integer string.
- _`mustBeIntInRange $min $max $value`_: Validates that `$value` is a base-10 integer string in the inclusive range `$min..$max`.

Example use:
```
{{ $value := somePipeline }}
{{ $allowedValues := list "foo" "bar" "baz" }}
{{ $sanitizedValue := $value | trim | mustBeOneOf $allowedValues }}
{{ /* the template rendering will fail at the above line if ($value | trim) does not evaluate as "foo", "bar" or "baz" /* }}
```

### DIFF
--- a/internal/template/validation_test.go
+++ b/internal/template/validation_test.go
@@ -1,0 +1,61 @@
+package template
+
+import "testing"
+
+func TestMustBeOneOf(t *testing.T) {
+	testCases := []struct {
+		name          string
+		allowedValues []any
+		input         string
+		wantValue     string
+		wantErr       bool
+		errSnippet    string
+	}{
+		{
+			name:          "valid input value",
+			allowedValues: []any{"a", "b", "c"},
+			input:         "a",
+			wantValue:     "a",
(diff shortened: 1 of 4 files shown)


internal/template/validation_test.go
```diff
@@ -0,0 +1,235 @@
+package template
+
+import (
+	"strings"
+	"testing"
+)
+
+func TestMustOneOf(t *testing.T) {
```
buchdag:
```suggestion
func TestMustBeOneOf(t *testing.T) {
```

